PR #67, opened 2022-08-31
NH-9823: Adding metric 'kube_node_spec_unschedulable'

gantrior:
This metric is indicating whether it is possible to schedule something on the node (useful for Healthscore)


pstranak-sw:
As I understand it, marking a Node as unschedulable is a user choice, like shutting it down. Should it decrease its healthscore?

pstranak-sw (approved):

PR #68, opened 2022-09-05
NH-9828: Adding container status metrics

gantrior:
* `k8s.container.status` - indicating status of the container
    * **approach**: combining 1 value datapoints of `kube_pod_container_status_waiting`, `kube_pod_container_status_running` and `kube_pod_container_status_terminated` metrics into single metric, and adding labels indicating the status value
* `k8s.container.ready` - indicating readyness of the container
    * **approach**: split `kube_pod_container_status_ready` into two metrics, one with datapoints indicating true, other indicating false values. Then combine them back which adds labels on datapoints based on its value

### DIFF
--- a/src/processor/swmetricstransformprocessor/factory.go
+++ b/src/processor/swmetricstransformprocessor/factory.go
@@ -122,9 +122,6 @@ func validateConfiguration(config *Config) error {
 			if op.Action == AddLabel && op.NewValue == "" {
 				return fmt.Errorf("operation %v: missing required field %q while %q is %v", i+1, NewValueFieldName, ActionFieldName, AddLabel)
 			}
-			if op.Action == FilterDataPoints && op.DataPointValue == 0 {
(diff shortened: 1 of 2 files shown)


src/processor/swmetricstransformprocessor/factory.go
```diff
@@ -122,9 +122,6 @@ func validateConfiguration(config *Config) error {
 			if op.Action == AddLabel && op.NewValue == "" {
 				return fmt.Errorf("operation %v: missing required field %q while %q is %v", i+1, NewValueFieldName, ActionFieldName, AddLabel)
 			}
-			if op.Action == FilterDataPoints && op.DataPointValue == 0 {
```
gantrior, line 125: this line is causing that we can't filter 0 value datapoint, which I think is legitimate use case.

juraj-paska (approved):

PR #72, opened 2022-09-20
NH-21693: Convert Cluster Status composite metrics to k8s Agent metrics

pstranak-sw:


deploy/k8s/manifest.yaml
```diff
@@ -335,6 +335,41 @@ data:
...
+          - include: k8s.node.status.condition.ready
+            action: insert
+            new_name: k8s.cluster.status.condition.ready
+            operations:
+              - action: aggregate_labels
+                label_set: []
+                aggregation_type: mean
+          - include: k8s.node.status.condition.networkunavailable
+            action: insert
+            new_name: k8s.cluster.status.condition.networkunavailable
```
pstranak-sw, line 347: I kept the names unchanged. Should we maybe add some suffix, like "avg"?

gantrior:
Hold on with this, I created https://swicloud.atlassian.net/browse/NH-22038 to update HealthScore calculation based on last discussion with Vish. We are going to remove conditions from HealthScore calculation

PR #79, opened 2022-09-30
NH-21202: Handling deployment condition

gantrior:
it is related to this PR: https://github.com/solarwindscloud/entity-service-schemas/pull/418

full list of conditions; https://github.com/kubernetes/api/blob/master/apps/v1/types.go#L494

### DIFF
--- a/deploy/k8s/manifest.yaml
+++ b/deploy/k8s/manifest.yaml
@@ -105,6 +105,22 @@ data:
               - action: update_label
                 label: status
                 new_label: sw.k8s.deployment.condition.progressing
+          - include: k8s.kube_deployment_status_condition
+            experimental_match_labels: { "condition": "ReplicaFailure" }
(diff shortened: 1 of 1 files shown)


deploy/k8s/manifest.yaml
```diff
@@ -73,6 +73,54 @@ data:
...
+                datapoint_value: 1
+                datapoint_value_action: include
+              - action: update_label
+                label: status
+                value_actions:
+                  - value: unknown
+                    new_value: "false"
+              - action: update_label
+                label: status
+                new_label: sw.k8s.deployment.condition.available
```
pstranak-sw, line 91: Since we are creating a new metric anyway, this rename is not really necessary. Unless we explicitly want to not use "status" to avoid confusion.

pstranak-sw (dismissed):

PR #81, opened 2022-10-04
NH-21704: Categorization of logs

gantrior:
adding `sw.k8s.log.type` attribute that indicates what is the type of the log:
* event - Kubernetes event logs
* journal - journal logs
* container - container logs


juraj-paska:
Journal logs do not need mapping to cluster?

juraj-paska (approved):

PR #89, opened 2022-10-17
NH-23685 Pod Detail Page - File system widgets

inforthecloud:
Creating metrics for Pod Filesystem widgets:

- _k8s.pod.fs.reads.rate_ and _k8s.pod.fs.writes.rate_ from **container_fs_reads_total** and **container_fs_writes_total**
- _k8s.pod.fs.reads.bytes.rate_ and _k8s.pod.fs.writes.bytes.rate_ from **container_fs_reads_bytes_total**  and **container_fs_writes_bytes_total**
- _k8s.pod.fs.usage_ from **container_fs_usage_bytes**

### DIFF
--- a/deploy/k8s/manifest.yaml
+++ b/deploy/k8s/manifest.yaml
@@ -386,31 +386,34 @@ data:
           - include: k8s.container_fs_reads_total
             action: insert
             match_type: regexp
-            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's file system read ops
+            # non-empty container, datapoints of Pod's file system read ops
             experimental_match_labels: { "container": "(.|\\s)*\\S(.|\\s)*" }
             new_name: k8s.pod.fs.reads.rate
           - include: k8s.container_fs_writes_total
             action: insert
             match_type: regexp
-            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's file system write ops
+            # non-empty container, datapoints of Pod's file system write ops
             experimental_match_labels: { "container": "(.|\\s)*\\S(.|\\s)*" }
             new_name: k8s.pod.fs.writes.rate
           - include: k8s.container_fs_reads_bytes_total
             action: insert
             match_type: regexp
-            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's file system read bytes
+            # non-empty container, datapoints of Pod's file system read bytes
             experimental_match_labels: { "container": "(.|\\s)*\\S(.|\\s)*" }
             new_name: k8s.pod.fs.reads.bytes.rate
           - include: k8s.container_fs_writes_bytes_total
             action: insert
             match_type: regexp
-            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's file system written bytes
+            # non-empty container, datapoints of Pod's file system written bytes
             experimental_match_labels: { "container": "(.|\\s)*\\S(.|\\s)*" }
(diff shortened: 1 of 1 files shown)


deploy/k8s/manifest.yaml
```diff
@@ -375,7 +375,24 @@ data:
               - action: update_label
                 label: owner_name
                 new_label: deployment
-
+          - include: k8s.container_fs_(reads|writes)_total
+            action: insert
+            match_type: regexp
+            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's CPU usage
+            experimental_match_labels: { "image": "^$", "pod": "(.|\\s)*\\S(.|\\s)*", "namespace": "(.|\\s)*\\S(.|\\s)*" }
+            new_name: k8s.pod.fs.iops
```
pstranak-sw, line 383: What is this metric supposed to contain? If I read the code right, we take metrics `k8s.container_fs_reads_total` and `k8s.container_fs_writes_total`, filter them and rename them to `k8s.pod.fs.iops`. So that the new metric will have data points mixed from datapoints of the original metrics. Did you want to sum them together? Or name the new metric differently?

inforthecloud: Yes, the intention is to combine reads and writes into a single metric. I can see now, that's not what I am achieving here, as reads and writes are essentially just renamed and never summed up. Could you advice how to achieve (reads + writes)? The same intention was for reads and writes bytes total.

inforthecloud:
> Did you try it locally using `skaffold dev` and check if it produces expected outcome?

I couldn't figure out how to check what it is producing. Could you please point me to the right direction?

deploy/k8s/manifest.yaml
```diff
@@ -375,7 +383,35 @@ data:
...
+            action: insert
+            match_type: regexp
+            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's file system written bytes
+            experimental_match_labels: { "container": "(.|\\s)*\\S(.|\\s)*" }
+            new_name: k8s.pod.fs.writes.bytes.rate
+          - include: k8s.container_fs_usage_bytes
+            action: insert
+            match_type: regexp
+            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's file system usage in bytes
+            new_name: k8s.pod.fs.usage.bytes
```
gantrior, line 414: you should be also summing this for each pod. Check how @michalJakubis is doing that in his PR: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/97/files

gantrior (changes requested):

michalJakubis (dismissed):

deploy/k8s/manifest.yaml
```diff
@@ -375,7 +383,38 @@ data:
...
+            new_name: k8s.pod.fs.reads.bytes.rate
+          - include: k8s.container_fs_writes_bytes_total
+            action: insert
+            match_type: regexp
+            # non-empty container, datapoints of Pod's file system written bytes
+            experimental_match_labels: { "container": "(.|\\s)*\\S(.|\\s)*" }
+            new_name: k8s.pod.fs.writes.bytes.rate
+          - include: k8s.container_fs_usage_bytes
+            action: insert
+            match_type: regexp
```
gantrior, line 412: what is the reason of `match_type: regexp`?

inforthecloud: Fixed

gantrior (approved):

PR #91, opened 2022-10-18
NH-23686 Pod network metrics 

DavidKana:
Added metrics for pods and nodes

- k8s.pod.network.*
- k8s.node.network.*
- k8s.node.fs.*

[Log for node](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/files/9912167/node.txt)
[Log for pod](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/files/9912168/pods.txt)


deploy/k8s/manifest.yaml
```diff
@@ -472,6 +526,85 @@ data:
             experimental_match_labels: { "sw.k8s.pod.status": "Running" }
             action: insert
             new_name: k8s.pod.status.phase.running
+          - include: k8s.container_network_receive_bytes_total
+            action: insert
+            operations:
+              - action: aggregate_labels
+                label_set: [node, namespace]
+                aggregation_type: sum
+            new_name: k8s.node.network.bytes_received
```
gantrior, line 535: I believe this transformation is not correct (the same for all other network related metrics). This is what is wrong:
* you are grouping that by `namespace`, I believe namespace does not play any role when you want to have datapoint per node, so I would skip the namespace
* you can't simply sum all the datapoints from this metric grouped by node. cAdvisor takes the metrics from cgroups and the structure of cgroups data looks like a tree, so you have datapoints for node, for pods and for containers, where the value of each cgroup is a cumulative for all its children. So if you sum that all you will have end result at least doubled. The solution is to take just datapoints for containers (using `container!=""` filter) and sum those. This is mentioned in disk Pod related metrics https://swicloud.atlassian.net/browse/NH-23685 (for node we just aggregate pod related metrics), and unfortunately for network metrics it is the same thing. Check PR of @inforthecloud https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/89/files

deploy/k8s/manifest.yaml
```diff
@@ -363,15 +375,57 @@ data:
...
             experimental_match_labels: { "image": "^$", "pod": "(.|\\s)*\\S(.|\\s)*", "namespace": "(.|\\s)*\\S(.|\\s)*" }
             new_name: k8s.pod.spec.cpu.quota
           - include: k8s.container_spec_cpu_period
             action: insert
             match_type: regexp
-            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's Memory usage
+            # empty `image` label and non-empty `pod` and `namespace` are datapoints of Pod's CPU period
             experimental_match_labels: { "image": "^$", "pod": "(.|\\s)*\\S(.|\\s)*", "namespace": "(.|\\s)*\\S(.|\\s)*" }
             new_name: k8s.pod.spec.cpu.period
+          - include: k8s.container_network_receive_bytes_total
```
gantrior, line 387: the same for my second point in other comment apply here (however grouping is correct here)

gantrior (changes requested):

deploy/k8s/manifest.yaml
DavidKana: Modified as suggested.

gantrior (approved):

gantrior (approved):

PR #92, opened 2022-10-19
Run memory_limiter as the first processor

pstranak-sw:
The `memory_limiter` should be the first processor to run in the pipeline (https://github.com/open-telemetry/opentelemetry-collector/blob/main/processor/memorylimiterprocessor/README.md#memory-limiter-processor) - moving it to the top of the list. Also added the limiter for log collectors.


juraj-paska (approved):

gantrior:
What about updating manifest version?

PR #94, opened 2022-10-20
NH-23684: Create Pod Memory / CPU utilization metrics

michalJakubis:
New metrics:

-  k8s.pod.spec.cpu.requests
-  k8s.pod.spec.memory.limit
-  k8s.pod.spec.memory.requests
-  k8s.pod.spec.cpu.limit


gantrior:
What about increasing manifest version?

juraj-paska (approved):

PR #103, opened 2022-11-14
Kubernetes node metrics calculation

DavidKana:
Modified `k8s.node.*` metrics calculations.
Added `k8s.node.fs.iops` and `k8s.node.fs.throughput` metrics


juraj-paska (commented):
Do we use all these metrics in SWO? These which are not used in SWO, should have sufix `_temp`. They are removed and not send to SWO.

DavidKana:
Unused metrics marked as _temp.

juraj-paska (approved):

PR #104, opened 2022-11-22
Update README with recent changes and support for Skaffold v2

pstranak-sw:


skaffold.yaml
```diff
@@ -3,40 +3,38 @@ kind: Config
...
-    - image: swi-k8s-opentelemetry-collector
-      docker:
-        dockerfile: build/docker/Dockerfile
 deploy:
   helm:
     releases:
-      - name: monitoring
-        remoteChart: prometheus
-        imageStrategy:
-          fqn: {}
```
pstranak-sw, line 18: This is the default value, so it's not necessary.

juraj-paska (dismissed):

skaffold2.yaml
```diff
@@ -0,0 +1,51 @@
...
+      namespace: monitoring
+      createNamespace: true
+      repo: https://prometheus-community.github.io/helm-charts
+  kustomize:
+    paths:
+    - tests/deploy/base
+    buildArgs:
+    - --load-restrictor LoadRestrictionsNone
+deploy:
+  kubectl: {}
```
pstranak-sw, line 26: When using both `helm` and `kustomize` manifests, we have to explicitly declare using `kubectl` and `helm` as the deployers. `kubectl` is used automatically only if there is no other deployer defined.

gantrior (approved):
Looks good

pstranak-sw:
Rebased on https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/105 to fix the build checks.

juraj-paska (approved):

PR #105, opened 2022-11-23
Fixing build

gantrior:
in new version of Helm chart there is probably no component called `node-exporter` so no events are triggered

### DIFF
--- a/.github/workflows/buildAndDeploy.yml
+++ b/.github/workflows/buildAndDeploy.yml
@@ -48,7 +48,7 @@ jobs:
 
       - name: Tail Otel mock events for 60 seconds
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/buildAndDeploy.yml
```diff
@@ -48,7 +48,7 @@ jobs:
 
       - name: Tail Otel mock events for 60 seconds
         run: |
-          kubectl delete pods -l component=node-exporter -n monitoring
+          kubectl run dummy --image ubuntu -- /bin/bash -ec "while :; do echo '.'; sleep 5 ; done"
```
pstranak-sw, line 51: Does it need to download the ubuntu image? Or is it already cached from some previous steps?

pstranak-sw (approved):

PR #106, opened 2022-11-25
NH-25910: Fixing memory leak

gantrior:
* updated builder golang image to 1.19.3
* updating OTEL collector to 0.64.0
* replacing `prometheustypeconverterprocessor` with `transformprocessor`


build/docker/Dockerfile
```diff
@@ -1,16 +1,17 @@
...
 
-RUN GO111MODULE=on go install go.opentelemetry.io/collector/cmd/builder@v0.61.0
+RUN GO111MODULE=on go install go.opentelemetry.io/collector/cmd/builder@v0.64.0
 
 WORKDIR /src
 COPY ["./src/", "./src/"]
 RUN CGO_ENABLED=0 /go/bin/builder --config ./swi-k8s-opentelemetry-collector.yaml --output-path ./
 
 # create vendor folder (for mend scanning)
+SHELL ["/bin/bash", "-c"]
```
pstranak-sw, line 14: Why this change?

build/swi-k8s-opentelemetry-collector.yaml
```diff
@@ -1,37 +1,34 @@
...
   - gomod: github.com/solarwinds/swi-k8s-opentelemetry-collector/processor/swmetricstransformprocessor v0.0.1
     path: "./src/processor/swmetricstransformprocessor"
 
 extensions:
-  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/healthcheckextension v0.61.0
-  - import: go.opentelemetry.io/collector/extension/ballastextension
-    gomod: go.opentelemetry.io/collector v0.61.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/healthcheckextension v0.64.0
+  - gomod: go.opentelemetry.io/collector/extension/ballastextension v0.64.0
+  - gomod: github.com/open-telemetry/opentelemetry-collector-contrib/extension/pprofextension v0.64.0
```
pstranak-sw, line 33: Do we want to have it available in the collector?

gantrior: I forgot to remove that. Fixed

build/docker/Dockerfile
gantrior: I had to adjust golang image from `alpine` to `bullseye`, because `alpine` does not have `git` in it and in new OTEL collector some packages are cloned directly from git, so it didn't work with `alpine`. And `bullseye` default shell is not bash, but `sh` so line 16 (`RUN if [[ -z "$CREATE_VENDOR_DIR" ]] ; then echo vendor creation skipped ; else go mod vendor ; fi`) did not work.

pstranak-sw (approved):

PR #115, opened 2023-01-05
NH-26016 - Create Error reason as mapping from combination of Event Reason and Type fields

etichy:
- Update otel collector to latest version 0.68
- Removed all deprecated code
- Update events processor to change severity according description (severity_number is set to lowest value from given [range ](https://github.com/open-telemetry/opentelemetry-specification/blob/main/specification/logs/data-model.md#field-severitynumber)13 - warning, 17 - error)

Testing:
 - create local cluster with failed image to deploy. events were send as error according specification
image

icon for buildin severity in log viewer cannot be changed as it's systelog severity threated as enum value, so we would need to map our number to syslog ....


build/docker/Dockerfile
```diff
@@ -4,7 +4,7 @@ FROM ${GOLANG_IMAGE} as builder
 
 COPY /build/swi-k8s-opentelemetry-collector.yaml /src/swi-k8s-opentelemetry-collector.yaml
 
-RUN GO111MODULE=on go install go.opentelemetry.io/collector/cmd/builder@v0.64.0
+RUN GO111MODULE=on go install go.opentelemetry.io/collector/cmd/builder@v0.67.0
```
gantrior, line 7: You should use the builder version aligned with processors, so please upgrade this to `0.68.0`. Build notifies you about that: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/actions/runs/3845213893/jobs/6549098289#step:5:383

gantrior (changes requested):

gantrior:
It would be probably better to create changes to the image in separate PR first, create docker hub image and then update manifest, because once merged master branch will contain manifest that won't be working, because it won't have the image upgraded yet.

But you can simply upgrade the image in this PR to new release, for example 0.2.0, and after this is merged you just trigger DockerHub deployment without need of extra PR.

build/docker/Dockerfile
etichy: I tired, but there's no build with version 0.68.0 :/

etichy:
> It would be probably better to create changes to the image in separate PR first, create docker hub image and then update manifest, because once merged master branch will contain manifest that won't be working, because it won't have the image upgraded yet.
> 
> But you can simply upgrade the image in this PR to new release, for example 0.2.0, and after this is merged you just trigger DockerHub deployment without need of extra PR.

done

gantrior:
> > It would be probably better to create changes to the image in separate PR first, create docker hub image and then update manifest, because once merged master branch will contain manifest that won't be working, because it won't have the image upgraded yet.
> > But you can simply upgrade the image in this PR to new release, for example 0.2.0, and after this is merged you just trigger DockerHub deployment without need of extra PR.
> 
> done

I meant [this version](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/Chart.yaml#L4) due to the fact that it is used as docker image tag [here](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/values.yaml#L16)

PR #117, opened 2023-01-05
Add support for HTTPS proxies

greg-poirier-swi:
The net/http package in Go supports the use of the HTTPS_PROXY environment variable to specify the URL of the proxy server. e.g. https://proxy.mydomain.com:8080. This exposes an otel.https_proxy_url value that users can use to set the environment variable.

### DIFF
--- a/deploy/helm/templates/common-env-config-map.yaml
+++ b/deploy/helm/templates/common-env-config-map.yaml
@@ -12,3 +12,6 @@ data:
   OTEL_ENVOY_ADDRESS_TLS_INSECURE: {{ quote .Values.otel.tls_insecure }}
   MANIFEST_VERSION: {{ quote .Chart.Version }}
   APP_VERSION: {{ quote .Chart.AppVersion }}
+{{ if .Values.https_proxy_url }}
+  HTTPS_PROXY: {{ quote .Values.https_proxy_url }}
+{{ end }}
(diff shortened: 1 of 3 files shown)


deploy/helm/templates/metrics-deployment.yaml
```diff
@@ -43,6 +43,8 @@ spec:
                   name: solarwinds-api-token
                   key: SOLARWINDS_API_TOKEN
                   optional: true
+            - name: HTTPS_PROXY
```
gantrior, line 46: I believe this is not right place to define this environment variable. It should be in https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/templates/common-env-config-map.yaml which is common for Otel collectors for metrics/events/logs, I believe this is applicable to all of them.

gantrior (changes requested):

deploy/helm/templates/metrics-deployment.yaml
greg-poirier-swi: You're right.

greg-poirier-swi:
We should really make it so that manifest.yaml just doesn't have checksums. That entire mechanism of deploying is going away, and it's tedious and unnecessary to have to update them.

greg-poirier-swi:
I'll squash merge and clean up the commit message. Where is the :shame: emoji.

gantrior (dismissed):

etichy (dismissed):

gantrior (approved):

PR #126, opened 2023-01-17
NH-29204: Adding semi-automated way of releasing Helm chart

gantrior:
* Used customized version of [chart-releaser-action](https://github.com/helm/chart-releaser-action/tree/v1.5.0) - mentioned the changes in license headers
* Making sure that produced release are not consider as DockerHub image releases
* Adjusted deployment documentation


mpecenka (approved):

pstranak-sw:
I think that instead of having a custom fork of the `chart-releaser-action`, we might as well write the script ourselves. We just need to:
1. Set the env variables
2. Download and extract the `cr` tool
3. Run `cr package ...`, `cr upload ...` and `cr index`

The code will be much more readable, maintainable (maybe 50 lines in total, instead of 350) and we will not have to deal with 3rd party license.

gantrior:
> I think that instead of having a custom fork of the `chart-releaser-action`, we might as well write the script ourselves. We just need to:
> 
> 1. Set the env variables
> 2. Download and extract the `cr` tool
> 3. Run `cr package ...`, `cr upload ...` and `cr index`
> 
> The code will be much more readable, maintainable (maybe 50 lines in total, instead of 350) and we will not have to deal with 3rd party license.
> 
> Edit: Maybe something like this: https://gist.github.com/pstranak-sw/958df33259c7e3a93063b528fdfc1bba (it could probably be even more simplified)

Ok it makes sense, it was not so hard to refactor it (thank you for providing the snippet). I updated PR (I verified the changes, so it works)

pstranak-sw (approved):

PR #137, opened 2023-01-23
NH-31513: Applying memory optimizations for Log collection

gantrior:
Explanation of the changes:
* Setting `filelog.storage`: This persist checkpoints in persistent storage, not in memory
* Setting `max_concurrent_files` to 10 (from default 1024), this is main memory optimization, reducing amount of concurrent log scans
* Increasing memory limit to `700Mi` (which should be enough for large logs)
* Having 150Mi difference between OTEL memory limit and Kubernetes memory limit, so that OTEL has enough buffer (this prevents OOMing)

Furthermore there was a bug when parsing time fraction, see https://stackoverflow.com/questions/53837975/time-parse-in-go-with-varying-fractional-second-length

### DIFF
--- a/deploy/helm/logs-collector-config.yaml
+++ b/deploy/helm/logs-collector-config.yaml
@@ -108,7 +108,7 @@ receivers:
     include_file_path: true
     include_file_name: false
     storage: file_storage
-    max_log_size: 10MiB
+    max_log_size: 1MiB
     max_concurrent_files: 10
     operators:
       # Find out which format is used by kubernetes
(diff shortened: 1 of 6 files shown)


deploy/helm/logs-collector-config.yaml
```diff
@@ -125,7 +130,7 @@ receivers:
         timestamp:
           parse_from: body.time
           layout_type: gotime
-          layout: '2006-01-02T15:04:05.000000000-07:00'
+          layout: '2006-01-02T15:04:05.999999999-07:00'
```
gantrior, line 133: Fixing how fraction of second is parsed, I found this issue during the analysis - see https://stackoverflow.com/questions/53837975/time-parse-in-go-with-varying-fractional-second-length for details

pstranak-sw (approved):

etichy (approved):

PR #151, opened 2023-02-01
Fix bugs with https_proxy_url and configmap checksums

greg-poirier-swi:


greg-poirier-swi:
Two commits:

    Properly annotate configmap checksums

    The configmap checksums should be on the resources created by
    DaemonSets, Deployments, etc--not on the DaemonSets, Deployments
    themselves. This moves the annotations to the pods for both the
    events and metrics collector and removes the unnecessary annotations
    on the logs daemonset.

    https_proxy_url is a top-level configuration value

    The value was originally placed under the otel scope within values.yaml,
    but referenced as .Values.https_proxy_url in the common-env ConfigMap
    template. Reconciling these differences by placing it at the top-level
    as it may be used in more places than just the collector.

gantrior:
Can we fix it in the template instead? https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/templates/common-env-config-map.yaml#L15. The value is configuring OTEL collector, I cannot think of any more `places` where it can be used

gantrior (approved):

PR #156, opened 2023-02-08
NH-31661: Adding integration tests for metric collection

gantrior:
* Adding integration tests in Python
* For integration tests the Prometheus is mocked with customized response
* Test checks whether outputs expected metric names and length of data (it is not possible to match content because even if I merge all JSON message into single json the order of resources is always different)


utils/cleanup_mocked_prometheus_response.py
```diff
@@ -0,0 +1,89 @@
...
+response = requests.get(url)
+if response.status_code == 200:
+    print("Successfully downloaded!")
+    lines = [line.decode() for line in response.content.splitlines()]
+    node = extract_property(lines, 'node')
+    apiserver_request_count = 0
+    result = [line for line in lines
```
juraj-paska, line 65: Some magic? :D Based on what this magic if happened? This would need to be edited if we poll new data, including the previous "url"?

tests/integration/README.md
```diff
@@ -0,0 +1,3 @@
+# Integration tests
```
juraj-paska, line 1: Should this be just empty?

tests/integration/test_metric_collection.py
```diff
@@ -0,0 +1,101 @@
...
+                        f'LineCount of outputs does not match, expected: {len(expected_json.splitlines())}, actual: {len(actual_json.splitlines())}')
+
+                metric_names = get_unique_metric_names(merged_json)
+                metric_matches = False
+                if all(name in metric_names for name in expected_metric_names):
+                    print("All specific metric names are found in the response.")
+                    metric_matches = True
+                else:
+                    print('Some specific metric names are not found in the response')
+                    print(f'Expected: {expected_metric_names}')
```
juraj-paska, line 62: Wouldn't this spam logs too much? Maybe just past what does miss.

tests/integration/test_metric_collection.py
```diff
@@ -0,0 +1,101 @@
...
+        else:
+            if response is not None:
+                print('Failed to download metrics. Response code:',
+                    response.status_code)
+            print('Retrying...')
+            time.sleep(2)
+
+    if time.time() - start_time >= timeout:
+        if not metric_matches and actual_json is not None:
+            print(f'Actual json:')
```
juraj-paska, line 83: Same here, Maybe just past what does miss.

utils/cleanup_mocked_prometheus_response.py
gantrior: I didn't want to write mock data manually from scratch, it would miss a lot of properties etc. So instead I took data from existing environment and filter it out. And this script is filtering this out. It needs to be run everytime we start scraping new data, and I documented the process of updating this here https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/b00b7159c52f54d63c108c7a72c080fb75910fd3/doc/development.md#updating-mocked-data-if-new-data-are-scraped 

But now I see I forgot to remove specific prometheushost that I used from code, I will fix that

tests/integration/test_metric_collection.py
```diff
@@ -0,0 +1,101 @@
...
+            print(actual_json)
+
+        raise ValueError("Timed out waiting for specific metric names")
```
juraj-paska, line 86: It would be maybe good to print these "indicators: processed_successfully and metric_matches and length_matches" here to simplify investigation.

tests/integration/README.md
gantrior: I removed the file

tests/integration/test_metric_collection.py
gantrior: Good point, adjusted

tests/integration/test_metric_collection.py
gantrior: added

juraj-paska (approved):

PR #174, opened 2023-02-23
NH-32001 Extend collector Healthcheck

DavidKana:
- Added Prometheus URI check using wget utility
- Added SWI OTEL endpoint check using 'fullstorydev/grpcurl' container and grpcurl utility
- Enabled 'check_collector_pipeline'

If above mentioned utilities fail, depending container is not started and problem is clearly indicated. 

Currently SWI OTEL endpoint doesn't support reflection, so required protofiles have been added.


deploy/helm/templates/logs-daemon-set.yaml
```diff
@@ -36,6 +36,38 @@ spec:
       tolerations:
         {{ toYaml . | nindent 8 }}
       {{- end }}
+      initContainers:
+        - name: otel-endpoint-check
+          image: fullstorydev/grpcurl
+          command: ['/bin/grpcurl', '-expand-headers',
+                    '-proto', 'opentelemetry/proto/collector/logs/v1/logs_service.proto',
+                    '-H', 'Authorization: Bearer ${SOLARWINDS_API_TOKEN}',
+                    '-d', '{ "resourceLogs": [ { "resource": { "attributes": [ { "key": "sw.k8s.log.type", "value": { "stringValue": "journal" } } ] } } ] }',
```
gantrior, line 45: can you add `sw.k8s.cluster.uid` resource attribute, so that the Log is assigned to the cluster and searchable through log explorer?

And what about adding there some log message?

deploy/helm/proto/collector/logs/v1/logs_service.proto
```diff
@@ -0,0 +1,79 @@
+// Copyright 2020, OpenTelemetry Authors
```
gantrior, line 1: instead of adding those files individually, can you set whole repo https://github.com/open-telemetry/opentelemetry-proto as submodule (it could be in the same folder `deploy/helm/proto`)

gantrior (changes requested):
Few comments. Also CHANGELOG update is missing

deploy/helm/proto/collector/logs/v1/logs_service.proto
DavidKana: Changed as suggested.

deploy/helm/templates/logs-daemon-set.yaml
DavidKana: Log message modified.

gantrior (approved):

deploy/helm/events-collector-config.yaml
```diff
@@ -6,7 +6,11 @@ exporters:
     headers:
       "Authorization": "Bearer ${SOLARWINDS_API_TOKEN}"
 extensions:
-  health_check: {}
+  health_check:
+    check_collector_pipeline:
+      enabled: true
```
pstranak-sw, line 11: When I was playing with this health check, I was not able to configure it work correctly. It kept failing after each pass of the `interval` and the collector was being regularly restarted. In the otel-contrib repo, it says that the extension works only with some exporters. Does it actually work with the  `otlp` exporter?

PR #176, opened 2023-02-28
NH-32156 Improve readme for SW Helm repo

juraj-paska:


README.md
```diff
@@ -1,7 +1,19 @@
 # SolarWinds Observability Kubernetes Helm Charts
 
+Kubernetes Helm Charts to monitor Kubernetes infrastructure using [SolarWinds Observability](https://documentation.solarwinds.com/en/success_center/observability/content/observability_documentation.htm).
```
pstranak-sw, line 3: Maybe we could use https://documentation.solarwinds.com/en/success_center/observability/default.htm#cshid=gh-k8s-collector instead?

juraj-paska: Well, the link you mentioned in used in "Usage" section. I meant this link to point to general documentation so customer can see what is SWO :)

gantrior (approved):

README.md
pstranak-sw: That's not the same link. It just happens to point to the same page for now. But the link that you added there is OK, too.

pstranak-sw (approved):

PR #184, opened 2023-03-06
NH-32001 initContainers health check - include opentelemetry protofiles when helm package is created

DavidKana:
When helm chart package is built, open telemetry proto files submodule should be included as well.
This PR also fixes local Skaffold deployment. 

- disabled `check_collector_pipeline`
- disabled otel endpoint and Prometheus in Skaffold 
- added .helmignore file to exclude unnecessary files from opentelemetry folder
- version updated to `2.2.0-alpha.3`

### DIFF
--- a/deploy/helm/.helmignore
+++ b/deploy/helm/.helmignore
@@ -1,0 +1,15 @@
+.helmignore
+
+.git
+.gitignore
+.gitattributes
+
+.github/
+.vscode/
+
+opentelemetry-proto/*.md
+opentelemetry-proto/*.yaml
+opentelemetry-proto/Makefile
+opentelemetry-proto/LICENSE
+
+
--- a/deploy/helm/CHANGELOG.md
+++ b/deploy/helm/CHANGELOG.md
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+## [2.2.0-alpha.3] - 2023-03-07
+
+### Added
+
+- Added opentelemetry protofiles
+
 ## [2.2.0-alpha.2] - 2023-03-02
 
 ### Added
--- a/deploy/helm/Chart.yaml
+++ b/deploy/helm/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: swo-k8s-collector
-version: 2.2.0-alpha.2
+version: 2.2.0-alpha.3
 appVersion: "0.3.0"
 description: SolarWinds Kubernetes Integration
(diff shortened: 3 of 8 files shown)


deploy/helm/.helmignore
```diff
@@ -0,0 +1,15 @@
...
+.gitattributes
+
+.github/
+.vscode/
+
+opentelemetry-proto/*.md
+opentelemetry-proto/*.yaml
+opentelemetry-proto/Makefile
+opentelemetry-proto/LICENSE
+
```
pstranak-sw, line 14: Personally, I'd exclude also protofiles, that we don't need. But I don't have a strong preference.

deploy/helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v1
 name: swo-k8s-collector
-version: 2.2.0-alpha.2
+version: 2.2.0-alpha.3
```
pstranak-sw, line 3: Let's revert this version change for now.

deploy/helm/CHANGELOG.md
```diff
@@ -9,6 +9,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+## [2.2.0-alpha.3] - 2023-03-07
```
pstranak-sw, line 12: Let's keep this as "Unreleased" for now. There are other changes, that need to be merged into this PR anyway.

deploy/helm/events-collector-config.yaml
```diff
@@ -8,7 +8,7 @@ exporters:
 extensions:
   health_check:
     check_collector_pipeline:
-      enabled: true
+      enabled: false
       interval: "5m"
       exporter_failure_threshold: 5
```
pstranak-sw, line 13: We can remove this whole section and revert it to the `health_check: {}` as it was originally.

deploy/helm/Chart.yaml
DavidKana: Reverted

deploy/helm/CHANGELOG.md
DavidKana: Removed

deploy/helm/events-collector-config.yaml
DavidKana: Removed as suggested.

deploy/helm/.helmignore
DavidKana: Let's keep these files for now.

pstranak-sw (approved):

PR #186, opened 2023-03-08
﻿NH-32185: Ignore datapoints from internal k8s containers (with container name "POD")

pstranak-sw:
We do not want to collect information about them nor create entities for them.


deploy/helm/metrics-collector-config.yaml
```diff
@@ -19,6 +19,20 @@ processors:
         statements:
           - convert_gauge_to_sum("cumulative", true) where IsMatch(metric.name, "^.*_total$") == true
 
+  # Remove datapoints of internal k8s containers from metrics, excluding the "container_network_.+_total" metric where other datapoints don't exist
```
pstranak-sw, line 22: While `container_fs_*`, `container_spec_*`, ... contain datapoints both for normal containers and the internal containers, `container_network_*` contain only datapoints for the internal containers. So we cannot get rid of them in the beginning of the pipeline - instead, we can do it only after we used them to create other metrics.

tests/integration/expected_metric_names.txt
```diff
@@ -9,7 +9,9 @@ k8s.cluster.nodes.ready
 k8s.cluster.nodes.ready.avg
 k8s.cluster.pods
 k8s.cluster.pods.running
+k8s.cluster.spec.cpu.requests
```
pstranak-sw, line 12: These metrics were missing in the test data. They are here now because of the changes in the `utils/cleanup_mocked_prometheus_response.py` script.

tests/integration/test_metric_collection.py
```diff
@@ -41,7 +42,7 @@ def assert_test_metric_names_found(content):
         metric_matches = True
     else:
         missing_metric_names = [
-            name in metric_names for name in expected_metric_names]
+            name for name in expected_metric_names if name not in metric_names]
```
pstranak-sw, line 45: A little bugfix to show the actual missing names.

tests/integration/test_utils.py
```diff
@@ -23,6 +23,7 @@ def retry_until_ok(url, func, print_failure):
     while time.time() - start_time < timeout:
         is_ok = False
         try: 
+            response = None
```
pstranak-sw, line 26: A bugfix to show a cleaner error message in case the request fails.

utils/cleanup_mocked_prometheus_response.py
```diff
@@ -16,46 +16,60 @@ def extract_property(array_of_lines, propertyName):
...
-        ('daemonset', 'test-daemonset'),
-        ('deployment', 'test-deployment'),
-        ('statefulset', 'test-statefulset'),
-        ('replicaset', 'test-replicaset'),
-        ('pod', 'test-pod'),
-        ('service', 'test-service'),
-        ('container', 'test-container'),
-        ('job', 'test-job'),
-        ('job_name', 'test-job-name'),
+    replacements: list[tuple[str, str, str | None]] = [
```
pstranak-sw, line 19: The whole script was changed to take the first datapoint for each found metric matching the set criteria. In addition to the previously existing criteria, it now ensures that if there are datapoints with attributes `container="POD"` and `container=".+"`, both datapoints are in the output dump. This is to enable a unit test in `tests/integration/test_metric_collection.py`.

gantrior (dismissed):

mpecenka (approved):

PR #190, opened 2023-03-09
NH-26100 - Container metrics

etichy:
Allow compute CPU Usage and CPU utilization on client. Fix container status to be placed in resource attribute so it's process correctly.


gantrior (dismissed):

gantrior:
You will need to walk through https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/doc/development.md#updating-tests-if-processing-is-changed to updated expected_output.json in integration test

gantrior (approved):

PR #196, opened 2023-03-13
NH-32808: Adding optional autoupdate of installed Helm chart

gantrior:


deploy/helm/templates/autoupdate/role.yaml
```diff
@@ -0,0 +1,16 @@
+{{- if .Values.autoupdate.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: Role
+metadata:
+  name: {{ include "common.fullname" . }}-autoupdate-role
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "common.labels" . | indent 4 }}
+rules:
+  - apiGroups:
```
jaroslav-fedor-swi, line 10: Can we limit these permissions to resource we know we touch?

gantrior: Yes it can be defined as:
```
- apiGroups:
  - ""
  resources:
  - pods
  - secrets
  - configmaps
  - services
  - serviceaccounts
  verbs:
  - get
  - list
  - create
  - delete
  - update
  - patch
- apiGroups:
  - batch
  resources:
  - jobs
  verbs:
  - get
  - list
  - create
  - delete
  - update
  - patch
- apiGroups:
  - apps
  resources:
  - deployments
  - daemonsets
  verbs:
  - get
  - list
  - create
  - delete
  - update
  - patch
- apiGroups:
  - rbac.authorization.k8s.io
  resources:
  - roles
  - rolebindings
  verbs:
  - get
  - list
  - create
  - delete
  - update
  - patch
```

but if we introduce new entities it wouldn't be able to cover that, and since this is Role so restricted to the namespace I thought it is OK to grant all permissions to that namespace. 

Do you prefer to list just specific resources?

gantrior: After offline discussion I adjusted PR with following:
* added comment above `autoupdate` to keep in mind that CronJob has full access to the namespace and is also able to modify ClusterRoles and ClusterRoleBindings
* added sha256 to the image

deploy/helm/values.yaml
```diff
@@ -288,6 +288,17 @@ cluster:
   name: <CLUSTER_NAME>
   uid: <CLUSTER_UID>
 
+# If enabled it creates CronJob that will periodically check for new versions of the Helm chart and upgrade if available
+# Keep in mind that in order to update resources the job has full access to the namespace where it is deployed and also have access to modify ClusterRole and ClusterRolBinding
```
pstranak-sw, line 292: typo in `ClusterRolBinding`

pstranak-sw (approved):

PR #201, opened 2023-03-14
Make integration tests more resilient against random ordering of elements and attributes in resulting OTEL messages

pstranak-sw:
The biggest change is in how the OTEL message for metrics is parsed and sorted so that the comparison with an expected message does not fail because of some random change in order of elements in the message.


tests/integration/test_utils.py
```diff
@@ -38,7 +38,7 @@ def retry_until_ok(url, func, print_failure):
                     print('Failed to download otel messages. Response code:',
                         response.status_code)
 
-                return False
```
pstranak-sw, line 41: When a network exception happened, the running test was marked as successful. This fixes it and allows to retry this step of the test.

tests/integration/test_metric_collection.py
```diff
@@ -11,19 +11,22 @@
...
+
+def test_expected_metric_names_are_generated():
     retry_until_ok(url, assert_test_metric_names_found,
                    print_failure_metric_names)
 
 
-def test_metric_line_count_generated():
-    retry_until_ok(url, assert_test_metrics_line_length_match,
-                   print_failure_line_count)
+def test_expected_otel_message_content_is_generated():
```
pstranak-sw, line 20: Changed name to better match the content of the test.

juraj-paska (approved):

gantrior (approved):

PR #209, opened 2023-03-17
NH-35868 - fix network/fs metrics on eks

etichy:
Fix grouping mechanism for container_network/fs metrics to depend only on existence of needed attributes rather then container attribute. 

+ Simplify regex to match non empty string

image


juraj-paska (approved):

deploy/helm/metrics-collector-config.yaml
```diff
@@ -494,7 +494,7 @@ processors:
         action: insert
         match_type: regexp
         # non-empty container, datapoints of Pod's file system write ops
```
pstranak-sw, line 496: The comments no longer match the filter.

pstranak-sw (approved):

gantrior (approved):

PR #210, opened 2023-03-17
NH-34013: Added automatic extraction of Kubernetes labels and annotations from resources

gantrior:


deploy/helm/metrics-collector-config.yaml
```diff
@@ -11,6 +11,55 @@ extensions:
 {{ toYaml .Values.otel.metrics.memory_ballast | indent 4 }}
 
 processors:
+{{- if .Values.otel.metrics.k8s_instrumentation.enabled }}
+  k8sattributes:
+    auth_type: "serviceAccount"
+    passthrough: false
+    extract:
+      metadata:
+        - k8s.deployment.name
```
etichy, line 20: Didn't we want to add those k8s attributes always so we can create direct link between deployment and pod? In case it's done like this we will lose that with disabled instrumentation

deploy/helm/metrics-collector-config.yaml
```diff
@@ -11,6 +11,55 @@ extensions:
...
+        - key_regex: (.*)
+          tag_name: k8s.pod.annotations.$$1
+          from: pod
+        - key_regex: (.*)
+          tag_name: k8s.namespace.annotations.$$1
+          from: namespace
+{{- end }}
+{{- if .Values.otel.metrics.k8s_instrumentation.labels.enabled }}
+      labels:
+        - key_regex: (.*)
```
etichy, line 37: Do we really want to add annotations/labels to all metrics? Isn't that too much? What's benefit to have it on all instead of create some helper metric which has just annotation purpose?

gantrior: We need this on custom dashboards (and maybe also Alerts and metric explorer), being able to create custom widget, selecting for example `k8s.container_cpu_usage_seconds_total` and filter datapoints based on labels. So we must assign labels to all those metrics. I hope it won't be too much, in sample data it looks like only 10% of increase, but we will see when we test this in dev-ssp

skaffold.yaml
```diff
@@ -46,6 +46,13 @@ deploy:
           otel.logs.filter.exclude.record_attributes[0].value: "monitoring"
           cluster.name: cluster name
           cluster.uid: cluster-uid-123456789
+          # excluded annotations:
+          # * kubectl.kubernetes.io/last-applied-configuration - deployed by kubectl, contains full config for each resource
+          # * cni.projectcalico.org/ - deployed in CI system (k3s)
+          otel.metrics.k8s_instrumentation.annotations.excludePattern: "(kubectl\\.kubernetes\\.io/last-applied-configuration)|(cni\\.projectcalico\\.org/.*)"
```
pstranak-sw, line 52: Does it make sense to exclude these also when deployed to an actual cluster?

gantrior: I am not sure. @jaroslav-fedor-swi do you want us to collect all annotations, including `last-applied-configuration` that is set by `kubectl apply` and includes json of full object?

PR #229, opened 2023-03-23
Features/nh 34992 fill system with container name

inforthecloud:
Specify syslog attributes to highlight pod name and container name in logs.


deploy/helm/logs-collector-config.yaml
```diff
@@ -229,6 +260,7 @@ service:
 {{- end }}
         - groupbyattrs/all
         - resource/container
+        - transform/syslogify
```
gantrior, line 263: I would move this transform up in the pipeline before `groupbyattrs/all`, so that you don't need to work with resource attribute there. If you process it after the grouping the transform needs to iterate all resources groups and I wouldn't be sure if it will work correctly in that case, maybe it will but it is safer to move that transform as close to the receiver as possible, so right after memory_limiter and filter.

After that you just update line 99 to `- set(body, Concat([attributes["syslogsev"], attributes["rawtime"], attributes["k8s.container.name"], attributes["k8s.pod.name"], "-", "-", "-", body ], " "))`

inforthecloud: Moved after filter.

gantrior (approved):

PR #233, opened 2023-03-23
NH-36372 install SWO Agent by default

jaroslav-fedor-swi:


deploy/helm/templates/swo-agent-statefulset.yaml
```diff
@@ -0,0 +1,39 @@
...
+    spec:
+      securityContext: {}
+      containers:
+        - name: swo-agent 
+          image: {{ .Values.swoagent.image }}
+          env:
+            - name: UAMS_CLIENT_ID_OVERRIDE_SOURCE_NAME
+              value: {{ .Values.cluster.uid }}
+            - name: SWO_URL 
+              value: {{ trimPrefix "otel.collector" .Values.otel.endpoint }}
```
pstranak-sw, line 31: Why are we modifying the `.Values.otel.endpoint` with the `trimPrefix` function?

deploy/helm/values.yaml
```diff
@@ -340,3 +340,17 @@ prometheus:
     # (this matches the behavior of `kube-prometheus-stack` chart).
     persistentVolume:
       enabled: false
+
+swoagent:
+  # Whether the SWO Agent should be deployed as part of this chart.
+  # If not, integrations are not available.
+  enabled: true
```
pstranak-sw, line 347: I would disable it by default for now until we have an actual usage for it in the product and we are ready for customers to use it.

jaroslav-fedor-swi: We have already a usage for it right now - you can configure it anytime. We only wont show widgets for a while.

deploy/helm/templates/swo-agent-statefulset.yaml
jaroslav-fedor-swi: As we are suppling otel.collector.<swo_url> where swo url can be na-01.cloud.solarwinds.com or na-02.cloud.solarwinds.com or any other datacell. I am just reusing the same parameter, so we dont need to change wizard.

deploy/helm/templates/swo-agent-statefulset.yaml
```diff
@@ -0,0 +1,39 @@
...
+            - name: UAMS_ACCESS_TOKEN 
+              valueFrom:
+                secretKeyRef:
+                  name: solarwinds-api-token
+                  key: SOLARWINDS_API_TOKEN
```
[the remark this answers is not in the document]
jaroslav-fedor-swi: As you dont provide secret for this pod, it wont be functional at all. I believe it should not be optional in this case.

deploy/helm/templates/swo-agent-statefulset.yaml
gantrior: In our wizard we prefill .Values.otel.endpoint to following value: `otel-collector.dc-01.cloud.solarwinds.com:443`, so it won't work anyway, you must cover also `otel-collector`

deploy/helm/templates/swo-agent-statefulset.yaml
```diff
@@ -0,0 +1,39 @@
+{{- if .Values.swoagent.enabled }}
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ .Values.cluster.name | replace " " "-" }}-swo-agent
```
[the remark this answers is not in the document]
gantrior: Kubernetes naming convention has following rules (see https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-label-names)
* contain at most 63 characters
* contain only lowercase alphanumeric characters or '-'
* start with an alphabetic character
* end with an alphanumeric character

If you want to use name input by the users you would need to cover all these

deploy/helm/templates/swo-agent-statefulset.yaml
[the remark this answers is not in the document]
pstranak-sw: Then maybe we should remove it from other places ... Not in this PR, though.

deploy/helm/templates/swo-agent-statefulset.yaml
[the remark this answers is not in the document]
pstranak-sw: I still don't like this. It's not consistent. And can we trust that the cluster name is valid for this usage if we just replace a space with a dash? I'd rather fix the issue with the UAMS agent itself.

deploy/helm/templates/swo-agent-statefulset.yaml
jaroslav-fedor-swi: `otel-collector` is deprecated and we should not use it anyway. but agree, we need to support both.

deploy/helm/values.yaml
pstranak-sw: I think there are still some rough edges that should be solved (like https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/233#discussion_r1147609223), before we enable it by default.

deploy/helm/templates/swo-agent-statefulset.yaml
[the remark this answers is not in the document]
jaroslav-fedor-swi: I added some simple logic to use cluster name if fits k8s naming convention, if not it will use release name with suffix. It is not ideal but could work in this beta before we get proper validation of name in wizard.

gantrior (approved):

pstranak-sw (dismissed):

pstranak-sw (approved):

PR #240, opened 2023-03-29
NH-34014: Scraping labels/annotations out of Deployments

gantrior:
NOTE: This PR does not fulful whole story, it is just one part of it. I want to split that into multiple iterative PR's for better reviews.

In this PR:
* Initial fork of k8sattributes processor (https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/240/commits/4b513e5f6e4b4382f2797fee8e33d0f4e257a95a)
* Adjusted processor to scrape Deployment labels/annotations (https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/240/commits/68fcd3d8e94132fbe0d2ee1151bbd289437bb2ff)

This will not be applied in Helm chart as it would need to update image, so that will be done in followup PR's.


etichy:
To make it super clear, it would be better to do just fork of original k8satttribte and in next PR do changes into it. Here it's hard to compare what's original code and what's updated

pstranak-sw (approved):

PR #248, opened 2023-04-11
NH-34014: Adding k8s instrumentation to Events

gantrior:
* using k8sattributes in Events collector
* extending its resource limits due to that
* refactored Helm template so that the definition code is deduplicated


juraj-paska (approved):

deploy/helm/values.yaml
```diff
@@ -190,6 +190,27 @@ otel:
...
+        enabled: true
+
+        # Provide a regular expression pattern to exclude specific labels from instrumentation.
+        # Example: To exclude labels with 'internal' or 'private' in their names, use the following pattern:
+        # excludePattern: ".*internal.*|.*private.*"
+        excludePattern:
+
+      annotations:
+        # Set 'enabled' to true to instrument Kubernetes annotations.
+        enabled: false
```
juraj-paska, line 208: Is this disabled just for event or also for metrics?

juraj-paska, line 208: Is this disabled just for events or also for metrics?

gantrior: We need annotations just to display its list for entities in UI, it won't be used for filtering, so we don't need to send them with both events and metrics. So it is enabled by default on metrics and disabled on events (to reduce data).

PR #254, opened 2023-04-18
NH-34806: Bundling kube-state-metric

gantrior:
The problem being solved here is that some init container related metrics were added recently to kube-state-metrics, and for example ssp clusters (kube-state-metrics v2.2) still has `kube_pod_init_container_resource_limits_cpu_cores` and `kube_pod_init_container_resource_limits_memory_bytes`, but it is deprecated and in newer versions there are only `kube_pod_init_container_resource_limits`. Bundling kube-state-metrics so that metrics that are scraped are predictable. 
* removing `prometheus.io/scrape` annotations so that it is not discovered by other prometheuses in the cluster
* adding another scrape target to scrape new kube-state-metric
* removing `k8s.cluster.memory.utilization` and `k8s.cluster.cpu.utilization` as they were combining kube_ and container_ metrics (which is not allowed now due to different scrape targets). This rely on this PR: https://github.com/solarwindscloud/entity-service-schemas/pull/688
* adjusting list of metrics
* Updating integration tests so that there are two separate responses and also jsonmerge did not work as expected when merging multiple sent messages so integration tests needed a lot of code adjustments


build/docker/wiremockMappings/redirectPrometheus.json
```diff
@@ -1,7 +1,7 @@
 {
   "request": {
     "method": "GET",
-    "urlPathPattern": "/federate*"
+    "url": "/federate?match%5B%5D=container_cpu_usage_seconds_total&match%5B%5D=container_spec_cpu_quota&match%5B%5D=container_spec_cpu_period&match%5B%5D=container_memory_working_set_bytes&match%5B%5D=container_spec_memory_limit_bytes&match%5B%5D=container_cpu_cfs_throttled_periods_total&match%5B%5D=container_cpu_cfs_periods_total&match%5B%5D=container_fs_reads_total&match%5B%5D=container_fs_writes_total&match%5B%5D=container_fs_reads_bytes_total&match%5B%5D=container_fs_writes_bytes_total&match%5B%5D=container_fs_usage_bytes&match%5B%5D=container_network_receive_bytes_total&match%5B%5D=container_network_transmit_bytes_total&match%5B%5D=container_network_receive_packets_total&match%5B%5D=container_network_transmit_packets_total&match%5B%5D=container_network_receive_packets_dropped_total&match%5B%5D=container_network_transmit_packets_dropped_total&match%5B%5D=apiserver_request_total&match%5B%5D=%7B__name__%3D%22kubernetes_build_info%22%2C+job%3D~%22.%2Aapiserver.%2A%22%7D"
```
etichy, line 4: Is that all container_* metrics?

tests/integration/expected_output.json
```diff
@@ -29095,18 +29082,6 @@
...
-                  {
-                    "asDouble": 18.595786343082114,
-                    "timeUnixNano": "0"
-                  }
-                ]
-              },
-              "name": "k8s.cluster.memory.utilization",
-              "unit": "Percent"
-            },
             {
```
etichy, line 29085: Missing new metrics for init container

deploy/helm/metrics-collector-config.yaml
```diff
@@ -1320,6 +1292,22 @@ receivers:
               - "container_network_transmit_packets_total"
               - "container_network_receive_packets_dropped_total"
               - "container_network_transmit_packets_dropped_total"
+              - "apiserver_request_total"
+              - '{__name__="kubernetes_build_info", job=~".*apiserver.*"}'
+{{- if .Values.otel.metrics.extra_scrape_metrics }}
+{{ toYaml .Values.otel.metrics.extra_scrape_metrics | indent 14 }}
```
etichy, line 1298: Are we going to check, that in extra_scrape_metrics isn't defined KSM metric?

build/docker/wiremockMappings/redirectPrometheus.json
gantrior: all container_, kubernetes_ and apiserver metrics. I just took full request. This would make sure that if we update list of metrics scraped we will need to update the mock as well

deploy/helm/metrics-collector-config.yaml
gantrior: intended use of extra_scrape_metrics are metrics from prometheus that customer needs (not us), so this should be convering only prometheus scraping and I assume that if use would need some addition kube_ metrics he would have them in prometheus and add them here are normally

etichy (approved):

PR #305, opened 2023-06-06
NH-42434: Use specific versions of 3rd party docker images

pstranak-sw:
Updating docker images `solarwinds/swo-agent`, `busybox`, `fullstorydev/grpcurl` and `alpine/k8s` to latest available versions.

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -386,8 +386,8 @@ swoagent:
   enabled: false
   image:
     repository: solarwinds/swo-agent
-    tag: "latest"
-    pullPolicy: Always
+    tag: ""
+    pullPolicy: IfNotPresent
   resources:
     limits:
       memory: 800Mi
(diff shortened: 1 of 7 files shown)


mpecenka (approved):

gantrior (changes requested):

deploy/helm/values.yaml
```diff
@@ -386,8 +386,8 @@ swoagent:
   enabled: false
   image:
     repository: solarwinds/swo-agent
-    tag: "latest"
-    pullPolicy: Always
+    tag: "v2.3.16@sha256:c42c9ca134db54341dc88c3e0c35b1a04db4629b1d97bfde524dfc5d2b8722e5"
```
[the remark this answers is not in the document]
pstranak-sw: Good point. Changed.

deploy/helm/values.yaml
```diff
@@ -386,7 +386,7 @@ swoagent:
   enabled: false
   image:
     repository: solarwinds/swo-agent
-    tag: "latest"
+    tag: ""
     pullPolicy: Always
```
gantrior, line 390: Do we still need always? `IfNotPresent` should be sufficient now

pstranak-sw: It should. There is no harm in any of those options. I can change that, too, if you want.

gantrior (approved):

PR #319, opened 2023-06-29
NH-44056: Calculating `k8s.pod.spec.cpu.limit` from kube-state-metrics

gantrior:


tests/integration/expected_output.json
```diff
@@ -22608,12 +22711,45 @@
               "gauge": {
                 "dataPoints": [
                   {
-                    "asDouble": 1,
+                    "asDouble": 0.1,
```
etichy, line 22714: Why we have different value here? Is that regression?

gantrior: since `k8s.pod.spec.cpu.limit` is now calculated from different metric, it takes different value from mocked data, and the value there is not really matching reality. In reality cpu limit from container runtime (cAdvisor) and cpu limit from kubernetes perspective (kube-state-metrics) should be the same most of the time.

etichy: So it means that mocked data are not in sync, right?

etichy (approved):

PR #330, opened 2023-07-11
NH-44282 filter out namespace phase metrics with values other than 1

mpecenka:
namespace phase metrics with values other than 1 have to be filtered out so the entity service does not ingest wrong state


deploy/helm/metrics-collector-config.yaml
```diff
@@ -252,6 +252,12 @@ processors:
           - action: filter_datapoints
             datapoint_value: 0
             datapoint_value_action: exclude
+      - include: k8s.kube_namespace_status_phase
```
etichy, line 255: As we are trying to not modify original metrics. Can we just make conditional our attribute which control namespace status?

https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/8c0f4b80e52ac520b4b6d225f376e2714e968af3/deploy/helm/metrics-collector-config.yaml#L1117

pstranak-sw (approved):

PR #344, opened 2023-08-11
NH-35608: Journal logs are not collected

gantrior:


tests/integration/test_utils.py
```diff
@@ -12,7 +12,7 @@ def get_all_log_resources(log_bulk):
     return result
 
 def get_all_bodies(log_bulk):
-    result = [records["body"]["stringValue"]
+    result = [records["body"]["stringValue"] if "stringValue" in records["body"] else records["body"]
```
pstranak-sw, line 15: Why this change?

gantrior: Because journal logs started to work in GitHub actions pipeline so started to appear in logs.json in mock service and integration test script was failing with the message that there is no `stringValue` on `body`. I assumed that journal logs has slightly different format than container logs. This condition fix it.

pstranak-sw (approved):

PR #359, opened 2023-08-28
NH-53106: Handling `k8s.node.name` attributes from Metrics correctly

gantrior:
`k8s.node.name` attributes from metrics often includes host name and not actual node name (they differ in case of Fargate). So I am setting the same mechanism we already previously did for Events. To remove `k8s.node.name` attribute if it does not match actual node name. 

In the same time I added instrumentation of `k8s.node.name` which should cover cases when it does not exists.

### DIFF
--- a/tests/integration/test_utils.py
+++ b/tests/integration/test_utils.py
@@ -115,6 +115,26 @@ def sort_attributes(element):
         element["attributes"] = sorted(
             element["attributes"], key=lambda a: a["key"])
 
+def sanitize_attributes(element):
+    if "attributes" in element:
+        # Exclude node attributes as they comes from real cluster so it is not predictable in integrations test to match them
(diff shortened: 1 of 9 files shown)


pstranak-sw (approved):

tests/integration/test_utils.py
```diff
@@ -115,6 +115,26 @@ def sort_attributes(element):
         element["attributes"] = sorted(
             element["attributes"], key=lambda a: a["key"])
 
+def sanitize_attributes(element):
+    if "attributes" in element:
+        # Exclude node attributes as they comes from real cluster so it is not predictable in integrations test to match them
+        exclude_prefixes = ['k8s.node.annotations', 'k8s.node.labels']
```
pstranak-sw, line 121: I's not great, that we have to exclude these.

pstranak-sw (approved):

PR #378, opened 2023-09-18
Nh 54277 windows logs monitoring

etichy:


deploy/helm/templates/logs-daemon-set.yaml
```diff
@@ -60,8 +60,9 @@ spec:
                 values:
                 - fargate
       {{- end }}
-      {{- if .Values.otel.logs.nodeSelector }}
       nodeSelector:
+        kubernetes.io/os: linux
+      {{- if .Values.otel.logs.nodeSelector }}
```
pstranak-sw, line 65: Why this change? Also, isn't it already covered by the `nodeAffinity`?

deploy/helm/templates/logs-daemon-set-windows.yaml
```diff
@@ -0,0 +1,143 @@
...
+              readOnly: true
+            - mountPath: c:\conf
+              name: opentelemetry-collector-configmap
+              readOnly: true
+            - name: logcheckpoints
+              mountPath: c:{{ .Values.otel.logs.filestorage.directory }}
+      volumes:
+        - name: varlogpods
+          hostPath:
+            path: c:\var\log\pods
```
pstranak-sw, line 126: Do we actually need the `c:\` everywhere? Isn't `\` enough?

deploy/helm/templates/logs-daemon-set-windows.yaml
```diff
@@ -0,0 +1,143 @@
...
+        - name: varlogcontainers
+          hostPath:
+            path: c:\var\log\containers
+        - name: varlibdockercontainers
+          hostPath:
+            path: c:\var\lib\docker\containers
+        - name: logcheckpoints
+          hostPath:
+            path: c:{{ printf "%s/%s" .Values.otel.logs.filestorage.directory .Values.cluster.uid }}
```
pstranak-sw, line 135: We may need to sanitize the `.Values.cluster.uid` here. Windows file system is more strict about allowed path names and the `.Values.cluster.uid` can be set by customers to basically anything.

deploy/helm/tests/__snapshot__/logs-daemon-set-windows_test.yaml.snap
```diff
@@ -0,0 +1,106 @@
...
+          - mountPath: c:\var\log\containers
+            name: varlogcontainers
+            readOnly: true
+          - mountPath: c:\var\lib\docker\containers
+            name: varlibdockercontainers
+            readOnly: true
+          - mountPath: c:\conf
+            name: opentelemetry-collector-configmap
+            readOnly: true
+          - mountPath: c:/var/lib/swo/checkpoints
```
pstranak-sw, line 79: So, are we using forward slashes or backward ones?

etichy: It doesn't matter, both are functional.  I wanted to preserve one configuration, that's why I did not changed checkpoint option.

deploy/helm/templates/logs-daemon-set-windows.yaml
etichy: good point, UID is GUID by default but it can be customized by customer. Hash will be better

deploy/helm/templates/logs-daemon-set-windows.yaml
etichy: \ is enough

deploy/helm/templates/logs-daemon-set.yaml
etichy: Well windows node require nodeSelector and affinity wasn't enough for him, So I tried to be consistent.

etichy: Well, we played with this error https://karpenter.sh/v0.29/troubleshooting/#windows-pods-are-failing-with-failedcreatedpodsandbox and we tried to follow instructions with @jaroslav-fedor-swi  and the only solution was for us move from affinity to node selector. It's hard to say whether it's really needed, but without rights I cannot change AIM roles.

etichy: Jaro found information where it's explicitly mentioned nodeSelector and without nodeSelector is really did not work. 

https://docs.aws.amazon.com/eks/latest/userguide/windows-support.html#remove-windows-support-data-plane

pstranak-sw (approved):

PR #384, opened 2023-10-02
NH-61248 Add services metrics

DavidKana:
Added Kube Service state metrics: https://github.com/kubernetes/kube-state-metrics/blob/main/docs/service-metrics.md


etichy:
Update metrics snapshot so we verify that metric is populated and tested

etichy (approved):

etichy (approved):

PR #400, opened 2023-10-11
NH-57813: Adding wrapper to Helm chart

gantrior:


deploy/helm/templates/logs-daemon-set-windows.yaml
```diff
@@ -58,9 +58,12 @@ spec:
           image: "{{ .Values.otel.image.repository }}:{{ .Values.otel.image.tag | default .Chart.AppVersion }}-nanoserver-ltsc2022"
           imagePullPolicy: {{ .Values.otel.image.pullPolicy }}
           command:
+            - c:\wrapper.exe
```
etichy, line 61: What about situation where with new helm they use old image? Bizzard situation, but possible.

etichy (approved):

deploy/helm/templates/logs-daemon-set-windows.yaml
gantrior: Helm chart brings the image v0.8.5 which has wrapper in it, customer would need to manually override it with `values.yaml` and I don't know why anyone would do that... so hopefully that situation is not realistic

PR #420, opened 2023-10-26
NH-65449 Fix CVE-2023-45142 swi-opentelemetry-collector high vulnerability

DavidKana:
Updated `go.opentelemetry.io/collector/*`, `google.golang.org/grpc` and other packages


gantrior (changes requested):
You can't just blindly update dependencies to OTEL components. OTEL components must match version of hosted OTEL collector, so if we are on v0.81.0, components cannot take dependencies from 0.88.0. This can be done either by:
* Updating OTEL collector to newer versions (e.q. https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/336)
* Patch individual dependencies using replacements in OTEL Collector builder which will replace dependencies in all components that use it (e.q. https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/270)

gantrior (approved):

PR #425, opened 2023-10-31
NH-65449 Update appVersion

DavidKana:
Update appVersion - The follow-up of https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/420


gantrior (commented):
You should also update CHANGELOG and fix expected outcome of integration tests (as image version is part of telemetry)

gantrior (approved):

PR #430, opened 2023-11-02
NH-47317 - Remove prometheus dependency

etichy:


.github/workflows/releaseHelm.yaml
```diff
@@ -30,9 +30,6 @@ jobs:
         with:
           version: v3.9.2
 
-      - name: Add dependency chart repos
-        run: helm repo add prometheus-community https://prometheus-community.github.io/helm-charts
-
```
gantrior, line 35: we take kube-state-metrics from that repo, so I think it is still needed there

deploy/helm/templates/metrics-collector-env-config-map.yaml
```diff
@@ -7,15 +7,6 @@ metadata:
...
 {{ include "common.labels" . | indent 4 }}
 data:
-{{- if .Values.prometheus.enabled }}
-{{- if .Values.prometheus.forceNamespace }}
-  PROMETHEUS_URL: "{{ .Release.Name }}-prometheus-server.{{ .Values.prometheus.forceNamespace }}.svc.cluster.local:80"
-{{- else }}
-  PROMETHEUS_URL: "{{ .Release.Name }}-prometheus-server.{{ .Release.Namespace }}.svc.cluster.local:80"
-{{- end }}
-{{- else }}
-  PROMETHEUS_URL: {{ quote .Values.otel.metrics.prometheus.url }}
```
gantrior, line 17: This should remain here. If `.Values.otel.metrics.prometheus.url` is filled with customer's deployed prometheus, we can scrape it. There is functionality in OTEL Collector for it: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/metrics-collector-config.yaml#L1504

gantrior (changes requested):
see my comments

.github/workflows/releaseHelm.yaml
gantrior: fixed

deploy/helm/templates/metrics-collector-env-config-map.yaml
gantrior: fixed

gantrior (approved):

PR #435, opened 2023-11-03
NH-64636: Decreased the default batch size

pstranak-sw:


gantrior (approved):
Unfortunatelly I just merge `alpha.3` just before few minutes, so I updated your PR to be `alpha.4`

PR #443, opened 2023-11-07
NH-63239: correctly filtering out `ebpf_net` metrics by default

gantrior:
* correctly filtering out ebpf_net metrics (those are internal metrics that customers does not need and can be opt-in via values.yaml)
* prefixing all metrics that arrives to OTEL endpoint with `k8s`, not just ebpf ones (for now it won't have impact, but in case customers would use that endpoint we would prefix it as well)

### DIFF
--- a/deploy/helm/metrics-collector-config.yaml
+++ b/deploy/helm/metrics-collector-config.yaml
@@ -171,13 +171,13 @@ processors:
         action: update
         new_name: k8s.$${1}$${2}
   
-  metricstransform/rename-ebpf:
+  metricstransform/rename-otel:
     transforms:
-      # add `k8s.` suffix to all metrics that are clearly provided by eBPF networking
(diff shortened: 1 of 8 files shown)


deploy/helm/metrics-collector-config.yaml
```diff
@@ -1690,10 +1690,10 @@ service:
         - otlp
       processors:
         - memory_limiter
-{{- if and .Values.ebpfNetworkMonitoring.enabled .Values.ebpfNetworkMonitoring.reducer.telemetry.metrics.enabled  }}
+{{- if and .Values.ebpfNetworkMonitoring.enabled (not .Values.ebpfNetworkMonitoring.reducer.telemetry.metrics.enabled)  }}
```
gantrior, line 1693: Here was wrong condition...

etichy (approved):

PR #445, opened 2023-11-07
NH-39459: Update integration tests to use real data and not mock of prometheus

gantrior:
* Updated test cases so that it tries to look for resource groups with given resource attributes and if it contains predefined metric name. The test also support asserting individual attributes of metric, but for this PR I am adding just basic tests
* Tests now heavily rely on deployment of test resources (covered in `test_metric_manifest.yaml`)
* Adding CI environment variable to be available in integration test run, so that tests can differentiate if they run in CI or locally


tests/integration/test_metric_collection.py
```diff
@@ -7,34 +7,236 @@
...
 
-# Commented till we resolve testing against real cluster
-# def test_expected_otel_message_content_is_generated():
-#     retry_until_ok(url, assert_test_expected_otel_message_content_is_generated,
-#                    print_failure_otel_content)
+    retry_until_ok(url, 
+                   lambda content: assert_metric_names_found(content, expected_metric_names),
+                   lambda content: print_failure_metric_names(content, expected_metric_names))
+    
+test_cases = [
```
etichy, line 28: These would be nice if it's split into separate files? Otherwise PR looks fine to me.

etichy (approved):

tests/integration/test_metric_collection.py
gantrior: Ok I will split that i followup PR, i will merge this to avoid nerge conflicts

PR #485, opened 2023-11-29
NH-64803 add opencost metrics

etichy:
Add opencost metrics collection, currently disabled and has to be enabled via values.yaml

### DIFF
--- a/deploy/helm/metrics-collector-config.yaml
+++ b/deploy/helm/metrics-collector-config.yaml
@@ -871,7 +871,18 @@ processors:
 
   batch:
 {{ toYaml .Values.otel.metrics.batch | indent 4 }}
-
+  filter/opencost-metrics:
+    metrics:
+      include:
+        match_type: regexp
+        metric_names:
+          - "^container_cpu_allocation$"
+          - "^container_gpu_allocation$"
+          - "^container_memory_allocation_bytes$"
+          - "^pod_pvc_allocation$"
+          - "^node_gpu_count$"
+          - "^kubecost_.*$"
+          - "^.*_cost$"
   filter/kube-state-metrics:
     metrics:
       include:
@@ -1004,6 +1015,21 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:{{ .Values.otel.metrics.otlp_endpoint.port }}
   k8s_events:
+{{ if .Values.opencost.enabled }}
+  prometheus/opencost:
+    config:
+      scrape_configs:
+        - job_name: opencost
+          honor_timestamps: false
+          honor_labels: true
+          scrape_interval: {{ quote .Values.otel.metrics.prometheus.scrape_interval }}
+          scrape_timeout: 10s
+          metrics_path: /metrics
+          scheme: http
+          static_configs:
+            - targets:
+              - ${OPENCOST_URL}
+{{- end }}
 {{- if and (and .Values.otel.metrics.extra_scrape_metrics (or (not .Values.otel.metrics.autodiscovery.prometheusEndpoints.enabled) .Values.otel.metrics.force_extra_scrape_metrics)) .Values.otel.metrics.prometheus.url }}
   prometheus/prometheus-server:
     config:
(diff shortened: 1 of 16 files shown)


deploy/helm/values.yaml
```diff
@@ -519,12 +519,40 @@ autoupdate:
...
   enabled: true
+  service:
+    annotations:       
+      opencost.scrape: "true"
+
+  prometheusScrape: false  
+
+  nodeSelector:
+    kubernetes.io/os: linux
+    kubernetes.io/arch: amd64
```
gantrior, line 530: please remove this constrain

deploy/helm/values.yaml
```diff
@@ -519,12 +519,40 @@ autoupdate:
...
+
+prometheus-node-exporter:  
+  service:
+    port: 9101
+    targetPort: 9101
+    annotations:
+      opencost.scrape: "true"     
+      prometheus.io/scrape: "false"
```
gantrior, line 538: Is this needed? isn't this covered by `prometheusScrape: false`?

deploy/helm/values.yaml
```diff
@@ -629,3 +657,197 @@ ebpfNetworkMonitoring:
       repository: ""
       tag: ""
       pullPolicy: IfNotPresent
+
+prometheus:
```
gantrior, line 661: I would explicitly comment here that this section is needed only when opencost is enabled otherwise is ignored.

deploy/helm/values.yaml
etichy: node exporter doesn't have such option

etichy: left over, remove that from values.yaml. Good catch

deploy/helm/values.yaml
```diff
@@ -519,12 +519,38 @@ autoupdate:
...
+prometheus-node-exporter:  
+  service:
+    port: 9101
+    targetPort: 9101
+    annotations:
+      opencost.scrape: "true"     
+      prometheus.io/scrape: "false"
+
+  nodeSelector:
+    kubernetes.io/os: linux
```
jaroslav-fedor-swi, line 541: Will it run on ARM as well?

deploy/helm/values.yaml
```diff
@@ -629,3 +655,196 @@ ebpfNetworkMonitoring:
...
+              values:
+              - fargate
+    nodeSelector:
+      kubernetes.io/os: linux
+
+opencost:
+  enabled: false
+  opencost:     
+    nodeSelector:
+      kubernetes.io/os: linux
```
jaroslav-fedor-swi, line 842: add arch here as well.

etichy: it supports ARM

deploy/helm/values.yaml
etichy: Yes, already checked

deploy/helm/metrics-collector-config.yaml
```diff
@@ -1004,6 +1023,21 @@ receivers:
       grpc:
         endpoint: 0.0.0.0:{{ .Values.otel.metrics.otlp_endpoint.port }}
   k8s_events:
+{{- if .Values.opencost.enabled }}
+  prometheus/opencost:
```
jaroslav-fedor-swi, line 1027: Does this prom "for opencost" needs to scrape opencost itself? Why?

deploy/helm/values.yaml
```diff
@@ -629,3 +655,196 @@ ebpfNetworkMonitoring:
       repository: ""
       tag: ""
       pullPolicy: IfNotPresent
+
+# Prometheus is deployed ony in case opencost section is enabled. Otherwise this section can be ignored.
```
jaroslav-fedor-swi, line 659:
```suggestion
# Prometheus is deployed only in case opencost section is enabled. Otherwise this section can be ignored.
```
typo

deploy/helm/metrics-collector-config.yaml
etichy: how would you scrape it? as application?

jaroslav-fedor-swi (approved):

gantrior (approved):

PR #511, opened 2023-12-12
EKS Add-on POC

jaroslav-fedor-swi:


gantrior (dismissed):
looks good

deploy/helm/templates/node-collector-daemon-set-windows.yaml
```diff
@@ -63,8 +63,8 @@ spec:
       {{- end }}
       containers:
         - name: swi-opentelemetry-collector
-          image: "{{ include "common.image" (tuple . .Values.otel "image_windows" nil (printf "%s%s" .Chart.AppVersion "-nanoserver")) }}"
-          imagePullPolicy: {{ .Values.otel.image_windows.pullPolicy }}
+          image: "{{ include "common.image" (tuple . .Values.otel.windows "image" nil (printf "%s%s" .Chart.AppVersion "-nanoserver")) }}"
```
gantrior, line 66: this is not good, `global.azure.images.image.*` is already reserved for linux image (those values are expected by AKS and they are flat). 

This can be solved by changing key just for Azure. You will need to update unit tests again though

```suggestion
          image: "{{ include "common.image" (tuple . .Values.otel.windows (tuple "image" "image_windows") nil (printf "%s%s" .Chart.AppVersion "-nanoserver")) }}"
```

gantrior (approved):

PR #513, opened 2023-12-13
NH-67718 Support pod monitor in k8s collector

jaroslav-fedor-swi:


deploy/helm/node-collector-config.yaml
```diff
@@ -452,11 +452,30 @@ receivers:
...
+      {{- if .Values.otel.metrics.autodiscovery.prometheusEndpoints.podMonitors }}
+      {{- range $key, $rule :=.Values.otel.metrics.autodiscovery.prometheusEndpoints.podMonitors.rules }}
+      prometheus/discovery/{{ $key }}:
+        rule: {{ $rule.rule }}
+        config:
+          config:
+            scrape_configs:
+              - job_name: pod
+                scheme: "http"
+                scrape_interval: {{ $rule.scrape_interval | default "60s" | quote}}
```
etichy, line 465: instead 60s would be nice to use .Values.otel.metrics.prometheus.scrape_interval

etichy (approved):

deploy/helm/node-collector-config.yaml
jaroslav-fedor-swi: I tried, but you cant use global inside loops.

PR #530, opened 2024-01-03
NH-69215:  Upgrade OTEL Collector and 3rd party Docker images

pstranak-sw:
Updating:
* OTEL Collector 0.81.0 -> 0.91.0
  * removed custom copy of `k8sconfig` package
  * backported all changes/fixes to `swmetricstransformprocessor`
  * backported all changes/fixes to `k8sattributesprocessor`
    * attribute `k8s.pod.start_time` now has a different format, which is technically a breaking change, but it's not used by SWO code
    * attribute `k8s.cluster.uid` is available only for Pods - it was not added to other entities handled by our custom code
    * config validation is implemented only for the root config - it was not added to other entities handled by our custom code
    * just like previously with ReplicaSet, the upstream code for supporting labels and annotations from Node was not backported as we prefer our custom version
  * unified `journald` receivers into one
    * there is no need to explicitly specify two log paths, as they are scanned by default (https://github.com/systemd/systemd/blob/42e6ad16843dc06a78e6db814e8a534aafc5139a/src/libsystemd/sd-journal/sd-journal.c#L1983-L1985)
    * specifying the paths stopped working for `journald` >= 0.82.0
* Docker images
  * `python:3.8-alpine` -> `python:3.12-alpine3.19`
  * `python:3.12-alpine3.17` -> `python:3.12-alpine3.19`
  * `debian:12.2` -> `debian:12.4`
  * `distroless/static` -> `distroless/static-debian12`
    * all other Debian based images were already using debian12
  * `golang:1.21.5-nanoserver-ltsc2022`
  * `alpine/k8s:1.27.2` -> `alpine/k8s:1.27.8`
  * `fullstorydev/grpcurl:v1.8.7` -> `fullstorydev/grpcurl:v1.8.9`
  * `bash:alpine3.16` -> `bash:alpine3.19`


src/processor/k8sattributesprocessor/options.go
```diff
@@ -37,6 +37,8 @@ const (
 	filterOPDoesNotExist = "does-not-exist"
 	metadataPodStartTime = "k8s.pod.start_time"
 	specPodHostName      = "k8s.pod.hostname"
+	// TODO: use k8s.cluster.uid from semconv when available, and replace clusterUID with conventions.AttributeClusterUid
```
etichy, line 40: Can you create JIRA and mention it here instead of TODO comment which will be forgotten? Thank

src/processor/k8sattributesprocessor/options.go
```diff
@@ -187,8 +189,8 @@ func withExtractMetadata(fields ...string) option {
 				p.rules.ContainerImageName = true
 			case conventions.AttributeContainerImageTag:
 				p.rules.ContainerImageTag = true
-			default:
-				return fmt.Errorf("\"%s\" is not a supported metadata field", field)
+			case clusterUID:
```
etichy, line 192: Missing default is from the original?

src/processor/k8sattributesprocessor/options.go
```diff
@@ -342,22 +344,16 @@ func withFilterLabels(filters ...FieldFilterConfig) option {
...
-				op = selection.Equals
 			case filterOPNotEquals:
 				op = selection.NotEquals
 			case filterOPExists:
 				op = selection.Exists
 			case filterOPDoesNotExist:
 				op = selection.DoesNotExist
 			default:
-				return fmt.Errorf("'%s' is not a valid label filter operation for key=%s, value=%s", f.Op, f.Key, f.Value)
+				op = selection.Equals
```
etichy, line 356: Default operators are from original?

etichy (approved):

deploy/helm/node-collector-config.yaml
```diff
@@ -294,14 +294,7 @@ connectors:
...
 {{- if and (not .isWindows) .Values.otel.logs.journal }}
-  journald/runlogs:
-    directory: /run/log/journal
-    units:
-      - kubelet
-      - docker
-      - containerd
-  journald/varlogs:
-    directory: /var/log/journal
+  journald:
```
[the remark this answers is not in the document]
etichy: Resolved by discussion: Would be nice to have and integration test which test journal log folder with default paths.

src/processor/k8sattributesprocessor/options.go
pstranak-sw: That's not our comment - it's from the upstream repo.

src/processor/k8sattributesprocessor/options.go
pstranak-sw: Yes. The error handling is now handled elsewhere.

src/processor/k8sattributesprocessor/options.go
pstranak-sw: Yes.

PR #535, opened 2024-01-08
Since costs were deprioritized - disable by default

jaroslav-fedor-swi:


pstranak-sw (approved):

pstranak-sw:
We could mention this change in the `/deploy/helm/CHANGELOG.md`. Though, in this case, it's not necessary.

PR #562, opened 2024-03-13
NH-73894 - document metrics endpoint

etichy:


README.md
```diff
@@ -113,3 +113,34 @@ otel:
           - key: k8s.namespace.name
             value: ^.*$
 ```
+
+## Receive 3d party metrics 
+
+SWO K8s Collector has otlp service endpoint which is able to receive and send metrics into SWO. All incoming metrics are decorated with prefix `k8s.` and properly associated with current cluster.
```
juraj-paska, line 119: > All incoming metrics
Shouldn't it be "All outgoing metrics" ?

juraj-paska (approved):

PR #561, opened 2024-03-13
NH-73933: Show an error when a deprecated OTEL endpoint is detected in `values.yaml`

pstranak-sw:
This is a follow-up to PR #563, failing the installation instead of just showing a warning. This should be merged in one of the future releases, ideally 3.4.0.

PR #563, opened 2024-03-13
NH-73933: Show warning when a deprecated OTEL endpoint is detected in `values.yaml`

pstranak-sw:
Similar to PR #561, just showing a warning instead of an error.


pstranak-sw:
I was unable to add unit test to https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/NH-73933-updateOtelEndpoint/deploy/helm/tests/notes_test.yaml for this. Anytime I try to add a new one, the already existing ones in the file start to fail for some formatting reasons.

gantrior (approved):

PR #575, opened 2024-04-10
Add JSON schema validation for the Helm chart

pstranak-sw:
The Helm chart now contains `values.schema.json`, which is used to validate user-provided configuration during installation/upgrade.

Summary of changes:
* Added schema to `deploy/helm/values.schema.json`.
  * Subchart sections are included, but only to the extend that we want customers to configure. E.g. container images used in them.
  * Schemas for k8s objects are included directly in the schema. Unfortunately, there is no official JSON schema that could be referenced, instead.
  * The schema checks for typos and unexpected settings/sections. :warning: This means that if there are invalid settings in the `values.yaml` provided by customers, the validation will fail with an error.
    For example, if there is a setting `something_unexpected: "whatever"` in the file, this will be the error:
    ```
    Error: INSTALLATION FAILED: values don't meet the specifications of the schema(s) in the following chart(s):
    swo-k8s-collector:
    - (root): Additional property something_unexpected is not allowed
    ```
* Adjusted the default `deploy/helm/values.yaml` to be compatible with the schema.
  * This affects only the default configuration for empty (not configured) settings. E.g. `additionalLabels:` -> `additionalLabels: {}`.
  * If users have `values.yaml` files with those settings explicitly mentioned but uninitialized, automatic merging of their settings with the default ones will ensure, that their `values.yaml` files will still pass the validation.
* Moved default values for referenced containers from templates to the `values.yaml`. It's more in line with where default values should be specified. It's also useful for documenting how to change all these images, if needed.

### DIFF
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -7,7 +7,7 @@ nameOverride: ""
 fullnameOverride: ""
 
 # array of pair "name: secret"
-imagePullSecrets:
+imagePullSecrets: []
 
 otel:
   # e.g. otel.collector.na-01.cloud.solarwinds.com:443
@@ -38,13 +38,13 @@ otel:
 
   init_images:
     swi_endpoint_check:
-      repository: ""
-      tag: ""
+      repository: "fullstorydev/grpcurl"
+      tag: "v1.8.9"
       pullPolicy: IfNotPresent
 
     busy_box:
-      repository: ""
-      tag: ""
+      repository: "busybox"
+      tag: "1.36.1"
       pullPolicy: IfNotPresent
 
   node_collector:
@@ -103,7 +103,7 @@ otel:
         #   `labels` - map of labels set on the pod
         #   `annotations` - map of annotations set on the pod
         # Example: namespace == "test-namespace" && labels["app"] == "test-app"
-        additionalRules:
+        additionalRules: ""
 
         podMonitors:
           rules: []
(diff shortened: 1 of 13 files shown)


deploy/helm/values.yaml
```diff
@@ -591,7 +592,7 @@ swoagent:
   enabled: false
   image:
     repository: solarwinds/swo-agent
-    tag: ""
+    tag: "v2.6.28"
```
gantrior, line 595: I am ok with defining tag in values for images like busybox etc, but I am not OK defining it for images which we need to upgrade regularly, which includes swo-agent. This will make very hard to upgrade those images with upgrade of helm charts, because many users simply copy whole `values.yaml` at the time of install, therefore they would be fixed to specific version of the image even after upgrade of Helm. 

> > Moved default values for referenced containers from templates to the values.yaml. It's more in line with where default values should be specified. It's also useful for documenting how to change all these images, if needed.

Do you have some reference material supporting this statement? What I see in e.g. Splung helm is that they use similar pattern for images they own (not all though): https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L957

gantrior (changes requested):

deploy/helm/values.yaml
pstranak-sw: This is a complicated topic.

From https://helm.sh/docs/chart_template_guide/functions_and_pipelines/#using-the-default-function:
> In an actual chart, all static default values should live in the values.yaml, and should not be repeated using the default command (otherwise they would be redundant). However, the default command is perfect for computed values, which cannot be declared inside values.yaml.

And Helm community has several discussions on this, e.g. `https://github.com/helm/helm/issues/7754`.
But the current "standard" seems to be that the main image should be using the `.Chart.AppVersion` by default, while 3rd party image tags are explicitly listed in `values.yaml`, sometimes as specific versions, sometimes as `latest`. As an example, here is such setting in the Splunk Helm, that you referenced: https://github.com/signalfx/splunk-otel-collector-chart/blob/main/helm-charts/splunk-otel-collector/values.yaml#L949.

Do we consider the UAMS agent a 3rd party image to the k8s collector?

And regarding copying the whole file by customers and using it just a bit modified, well, that can happen. But does it really? The file is getting bigger and bigger. And now with schema validation, they should have even fewer reasons to not just write their own overrides.

gantrior (approved):

deploy/helm/values.yaml
pstranak-sw: Yes, we used to have such concerns. But that was when the default file had just a few dozen lines, at most. Today, I doubt that anyone would just copy it.

etichy (approved):

PR #600, opened 2024-04-18
Update stanza package to restore a behavior from before update to 0.98.0

pstranak-sw:
Update of OTEL collector to 0.98.0 in PR #579 introduced a change in how logs were recombined. Update the `stanza` package to the version, in which it allows to restore previous behavior. There is no other change in the package since 0.98.0.


gantrior:
You can't update Helm and Docker image in the same PR. It will turn Helm chart into broken state, as it will use config on old image. This is what `helm_e2e` job is testing, if current OTEL config in Helm works with docker image pulled used in Helm (so without building it). So you must first update docker image, publish it, and then use in Helm and update config.

gantrior:
If this works we can close the revert right? https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/599

pstranak-sw:
If make this and #601 into one, `helm_e2e` job will fail. If I keep them separated, `build_and_test` will fail. I guess I will have to disable the problematic test in this PR and reenable it in the other one.

gantrior (approved):

PR #603, opened 2024-04-19
NH-73418: Adding CLUSTER_UID to agent statefulset

gantrior:
There are 3 environment variables added which can be useful to the SWO Agent: 
* CLUSTER_UID
* CLUSTER_NAME
* MANIFEST_VERSION - Helm chart version


deploy/helm/templates/swo-agent-statefulset.yaml
```diff
@@ -50,6 +50,9 @@ spec:
                 secretKeyRef:
                   name: {{ template "common.secret" . }}
                   key: SOLARWINDS_API_TOKEN
+          envFrom:
+            - configMapRef:
+                name: {{ include "common.fullname" (tuple . "-common-env") }}
```
pstranak-sw, line 55: This will add also variables like `HTTPS_PROXY`, which could have other meaning for the SWO Agent. Is that expected?

gantrior: removed, discussed offline

pstranak-sw (approved):

PR #623, opened 2024-05-07
NH-72863 cluster version fix

mpecenka:
fixed by creating sw.k8s.cluster.version on k8s.cluster.version metric only


pstranak-sw:
Don't forget to mention this in the changelog.

pstranak-sw (approved):

PR #632, opened 2024-05-24
NH-36351: Removed use of `swmetricstransform`

gantrior:
* This processor is no longer needed since filterprocessor can filter by datapoint values
* `value_int` does not work, all values in our pipeline are `value_double`
* filtering is done in separate step of the pipeline.
* Verified manually. 
* I change both (image and helm) in the same time, because Helm does not require new updated image.


deploy/helm/node-collector-config.yaml
```diff
@@ -75,6 +75,12 @@ processors:
     transforms:
       {{- include "common-config.metricstransform-preprocessing-cadvisor" . | nindent 6 }}
 
+  filter/preprocessing-cadvisor:
+    error_mode: ignore
+    metrics:
+      datapoint:
+        - 'metric.name == "k8s.container.status" and value_double != 1'
```
pstranak-sw, line 82: Why? Metric `k8s.container.status` is created only in the metrics-collector. It does not exist in the node-collector, so there is nothing to filter.

pstranak-sw:
Could you add a note to the unreleased section of the chart's changelog that we are replacing all usages of `swmetricstransformprocessor` for `filterprocessor`?

deploy/helm/node-collector-config.yaml
gantrior: Ah, I thought it is coming from cAdvisor which can be scraped on both. Ok removed from node-collector and merged `filter/preprocessing-cadvisor` with `filter/preprocessing`

pstranak-sw (approved):

pstranak-sw (approved):

PR #634, opened 2024-05-24
NH-80889: Renaming k8sattributes to swk8sattributes

gantrior:
* Renamed module name to `github.com/solarwinds/swi-k8s-opentelemetry-collector/processor/swk8sattributesprocessor` 
    * after doing that I had to pull `github.com/solarwinds/swi-k8s-opentelemetry-collector/internal/k8sconfig` to the code base, because it became unavailable for the module (as it is internal)
* Renamed name of the processor to `swk8sattributes`
* Fixed all the usages in Helm
* Bumped docker image to 0.12.0
* Bumped Helm chart to 4.0.0-alpha.6

The build won't pass, but this is so breaking that I can't even change image first, so changing both Helm and image at once and I will publish docker image as soon as this get merged to master.


src/cmd/main.go
```diff
@@ -33,7 +33,7 @@ func main() {
 	info := component.BuildInfo{
 		Command:     "swi-k8s-opentelemetry-collector",
 		Description: "SolarWinds distribution for OpenTelemetry",
-		Version:     "0.11.1",
+		Version:     "0.12.0",
```
pstranak-sw, line 36: Why not continue with 0.11.x? It shouldn't be used by any released `k8s collector` yet.

src/internal/k8sconfig/go.mod
```diff
@@ -0,0 +1,60 @@
+module github.com/solarwinds/swi-k8s-opentelemetry-collector/internal/k8sconfig
+
+go 1.21.0
```
pstranak-sw, line 3: We set version `1.22` in other go.mod files. Not that it makes much difference, anyway.

src/cmd/main.go
gantrior: Ok I updated image to 0.11.2.

src/internal/k8sconfig/go.mod
gantrior: Updated

pstranak-sw (dismissed):

pstranak-sw (approved):

PR #635, opened 2024-05-28
Add support for new filtering syntax

pstranak-sw:
A continuation of PR #628.

Restore backwards compatibility for `otel.metrics.filter`, `otel.logs.filter` and `otel.events.filter` (follow-up to 5b8d2c7)

The basic idea is that when users are setting the filters using old filtering syntax, we merge them with the default filters in the old syntax and use the old attributes behavior (they are in records).
And when they use the new syntax, we merge them with the default filters in the new syntax and use the new attributes behavior (they are in resources).

Because the new syntax and the old syntax are not compatible (they cannot be used both at the same time), and until now we forced users to use the old syntax (because our default filters were using the old syntax), we can use this as the "master" switch, allowing for backwards compatibility.

### DIFF
--- a/README.md
+++ b/README.md
@@ -109,7 +109,7 @@ otel:
     filter:
       include:
         match_type: regexp
-        resource_attributes:
+        record_attributes:
           - key: k8s.namespace.name
             value: ^.*$
 ```
--- a/deploy/helm/tests/__snapshot__/node-collector-config-map_test.yaml.snap
+++ b/deploy/helm/tests/__snapshot__/node-collector-config-map_test.yaml.snap
@@ -1,4 +1,1212 @@
-Node collector config should match snapshot when autodiscovery is disabled:
+Custom logs filter with new syntax:
+  1: |
+    common.proto: ""
+    logs.config: |
+      connectors:
+        forward/logs-exporter: null
+        forward/metric-exporter: null
+      exporters:
+        otlp:
+          endpoint: ${OTEL_ENVOY_ADDRESS}
+          headers:
+            Authorization: Bearer ${SOLARWINDS_API_TOKEN}
+          retry_on_failure:
+            enabled: true
+            initial_interval: 10s
+            max_elapsed_time: 300s
+            max_interval: 30s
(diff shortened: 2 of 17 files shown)


deploy/helm/values.yaml
```diff
@@ -185,12 +185,7 @@ otel:
 
     # This filter is applied after metric processing, it is the place where metrics could be filtered out
     # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor for configuration reference
-    filter:
-      exclude:
-        match_type: regexp
-        metric_names:
-          - .*_temp
-          - apiserver_request_total
```
pstranak-sw, line 193: Let's just hard-code this. There is no scenario where we want customers to change these default filters. That way we don't have to deal with default filters in old vs. new filtering syntax.

README.md
```diff
@@ -109,7 +109,7 @@ otel:
     filter:
       include:
         match_type: regexp
-        resource_attributes:
+        record_attributes:
```
pstranak-sw, line 112: This was changed in the previous PR. Reverting back, because the README is for the current stable release, which does still work with `record_attributes`.

deploy/helm/tests/__snapshot__/node-collector-config-map_test.yaml.snap
```diff
@@ -165,19 +2584,9 @@ Node collector config should match snapshot when autodiscovery is disabled:
...
-              - key: k8s.namespace.name
-                value: ^kube-.*$
-        filter/metrics:
-          metrics:
-            exclude:
-              match_type: regexp
-              metric_names:
-              - .*_temp
-              - apiserver_request_total
+          logs: |-
```
gantrior, line 2587: What does `|-` do there? Are you sure that this works?

pstranak-sw: No. I forgot a leftover debug code in there and it caused this. Fixed.

PR #646, opened 2024-06-23
NH-68385: Watching and Polling all supported resources and sending their manifest to SWO as logs.

gantrior:
This depends on https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/645

- Watching all supported resources and sending their full manifests to SWO.
  - Disabled by default, can be enabled by setting `otel.events.manifests.enabled` to true.
  - Runs in event collector, so require `otel.events.enabled` set to true. 
- Sends events to `sw.events.inframon.k8s.manifests` which is yet to be created.


deploy/helm/events-collector-config.yaml
```diff
@@ -166,6 +271,30 @@ service:
...
+{{- if .Values.otel.events.manifests.enabled }}
+    logs/manifests:
+      exporters:
+        - otlp
+      processors:
+        - memory_limiter
+        - transform/manifest
+        - groupbyattrs/manifest
+        - resource/manifest
+        - k8sattributes
```
pstranak-sw, line 284: Does it mean we will annotate the manifests with labels/annotations even though they are already part of the manifests?

deploy/helm/values.yaml
```diff
@@ -284,9 +284,21 @@ otel:
...
   events:
-    # Define whether events will be collected and sent
+    # If false none of the events will be collected
     enabled: true
 
+    # If true, the k8s events will be collected
+    k8s_events: true
+
+    # Configuration that allows to watch for changes in k8s resources and send full manifests to SWO upon change
+    manifests:
```
pstranak-sw, line 294: I get why you put this in the `events` section. But logically, there is no relation between collection of manifests and collection of events. They just happen to run in the same container. The current implementation also means that all filtering and k8s instrumentation settings must apply to both events and manifests equally.

deploy/helm/events-collector-config.yaml
gantrior: Yes it annotates with labels and annotations setting them to resource attributes, so that it is mapped to the entity. I don't know how to expand annotations/labels from manifest to resource attributes using OTTL. It is probably possible, but it was easier to leverage k8sattributes for this job.

deploy/helm/values.yaml
gantrior: Yes I set to pipeline to run in events collector and introduced manifests as second type of events. I turned k8s event collector to just event collector which scrape two types of events. It made sense to me as it is kind of event that we will send to another event namespace. Therefore `events.enabled` is turning on/off events collector, `events.k8s_events` is turning on/off k8s events collection and `events.manifests.enabled` is turning on/off manifests collection. Also since it all runs in the same deployment it logically share batching, queue configuration, timeouts, resource configuration etc. Having shared filter is not ideal though. It would be probably better to introduce new section for shared stuff, like `event_collector`, move there resource configuration, telemetry, batching, timeout etc.and then having manifests separate. What do you suggest?

gantrior: After discussion I moved `manifests` to its own section.

deploy/helm/CHANGELOG.md
```diff
@@ -7,6 +7,14 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+## [4.0.0-alpha.3] - 2024-06-25
+
+## Added
+
+- Watching all supported resources and sending their full manifests to SWO.
+  - Disabled by default, can be enabled by setting `otel.manifests.enabled` to true.
+  - Runs in event collector, so require `otel.events.enabled` set to true.
```
pstranak-sw, line 16: Later, we could change the chart to render the event's Deployment and ConfigMap when either `.Values.otel.events.enabled` or `.Values.otel.manifests.enabled` is set.

pstranak-sw (approved):

pstranak-sw (approved):

deploy/helm/CHANGELOG.md
```diff
@@ -465,7 +482,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
-- Scrape kube_service_* and kube_endpoint_* metrics
+- Scrape kube*service*_ and kube*endpoint*_ metrics
```
pstranak-sw, line 485: @gantrior You might want to adjust your IDE. You keep refactoring all markdown files you touch. And the automatic refactoring can cause errors like this.

gantrior: damn you are right

gantrior: I reverted those changes

pstranak-sw (approved):

PR #653, opened 2024-07-01
Merge `release/3.4.0` to `main`

pstranak-sw:
Merge changes from the `release/3.4.0` branch into the `main`.

PRs from the 3.4.0: #647, #648, and #650

Changed in addition to what is in 3.4.0:
* Since the image in `main` already contains additional changes, a new one will need to be released as `0.11.1`. This will be covered in a later PR.
* The main package of the k8s collector binary was updated to match the upstream OTEL Collector

TODO:
* Once this is merged, release image `0.11.1` and then release a new `k8s collector` alpha consuming it.


src/cmd/main.go
```diff
@@ -19,23 +19,49 @@ import (
...
+		BuildInfo: info,
+		Factories: components,
+		ConfigProviderSettings: otelcol.ConfigProviderSettings{
+			ResolverSettings: confmap.ResolverSettings{
+				ProviderFactories: []confmap.ProviderFactory{
+					envprovider.NewFactory(),
+					fileprovider.NewFactory(),
+					httpprovider.NewFactory(),
+					httpsprovider.NewFactory(),
+					yamlprovider.NewFactory(),
```
pstranak-sw, line 49: By default, the OTEL Collector supports getting a whole configuration or fragments of a configuration using all these providers. But we actually only use the `fileprovider` (for loading the configuration file itself) and the `envprovider` (for resolution of environment variables in the configuration). 

The `yamlprovider` could be used for command line overrides, which we don't currently use. And while the `httpprovider` and the `httpprovider` could be used for getting the configuration dynamically from our servers, we do not plat to use these for now.

We could remove the the last three if we wanted to prevent people from using them.

PR #665, opened 2024-07-11
NH-80891 - new filtering syntax

etichy:
Switch our otel config to use new OTTL filtering syntax on all places


README.md
```diff
@@ -101,17 +101,14 @@ Processors included in the collector:
 
 Once deployed to a Kubernetes cluster, the logs collection and processing configuration is stored as a ConfigMap under the `logs.config` key.
 
-To avoid processing an excessive amount of data, the `swo-k8s-collector` collects container logs only in `kube-*` namespaces, which means it only collects logs from the internal Kubernetes container. This behavior can be modified by setting `otel.logs.filter` value. An example for scraping logs from all namespaces:
+Collector by default gather all logs which migh be intensive. To avoid processing an excessive amount of data, the `swo-k8s-collector` can define filter which will drop all unwanted logs. This behavior can be modified by setting `otel.logs.filter` value. An example for scraping logs only from `kube-*` namespace:
```
pstranak-sw, line 104: I would rather not change this until we release 4.x. This is what the users of 3.x will read, too, and it might be confusing.
Or, maybe even better, describe both behaviors here, the one in 3.x and the change in the upcoming 4.x.

deploy/helm/tests/__snapshot__/metrics-collector-config-map-fargate_test.yaml.snap
```diff
@@ -507,99 +507,56 @@ Metrics config should match snapshot when using default values:
...
-              - ^kube_service_spec_external_ip$
-              - ^kube_service_status_load_balancer_ingress$
-              - ^kube_endpoint_address_not_ready$
-              - ^kube_endpoint_address_available$
-              - ^kube_endpoint_info$
-              - ^kube_endpoint_created$
-              - ^kube_endpoint_ports$
-              - ^kube_endpoint_address$
+            metric:
+            - "not(\nname == \"kube_deployment_created\" or\nname == \"kube_deployment_created\"
```
pstranak-sw, line 519: It's a bit unfortunate, that using `IsMatch` in the condition causes this unreadable mess in the rendered file. But I don't see a way around it.

pstranak-sw:
Please mention the change in the Chart's changelog.

deploy/helm/tests/__snapshot__/metrics-collector-config-map-fargate_test.yaml.snap
etichy: workaround is to use > instead of | , where everything will be one giant line condition if it's better

pstranak-sw (approved):

PR #679, opened 2024-08-14
Upgrade kube-state-metrics and OTEL Collector

pstranak-sw:
- Upgraded the `kube-state-metrics` subchart from `5.15.2` to [5.25.1](https://github.com/prometheus-community/helm-charts/releases/tag/kube-state-metrics-5.25.1) which brought the following changes:
  - Upgraded the `kube-state-metrics` image to [2.13.0](https://github.com/kubernetes/kube-state-metrics/releases/tag/v2.13.0).
  - Added `restart_policy` to the `kube_pod_init_container_info` metric.
  - Added new metric `kube_pod_container_status_last_terminated_timestamp`.
  - For the list of all changes see the respective Git repositories. (they should not have any effect on the `k8s collector`)
- Upgraded collector image to `0.11.4` which brings following changes:
  - Upgraded OTEL Collector to v0.107.0.
  - Removed deprecated `ballastextension`, because it wasn't working with the `memorylimiterprocessor` anymore. The collector now uses environment variable `GOMEMLIMIT`. The variable is taken from the container resource limits, if they are set. If they are not set, Kubernetes fills this value with Node allocatable.
  - Updated Go builder image
- Fixed typo in filter syntax warning
- Upgraded SWO Agent image to `v2.9.3`

There will be a separate PR for actually using the new image in the Helm chart.


deploy/helm/templates/events-collector-deployment.yaml
```diff
@@ -83,6 +83,10 @@ spec:
                   name: {{ template "common.secret" . }}
                   key: SOLARWINDS_API_TOKEN
                   optional: true
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
```
pstranak-sw, line 89: This uses the configured memory limits of the Container. If they are not configured, Kubernetes fills this value with Node allocatable.

etichy (approved):
🆗

PR #683, opened 2024-08-18
Adding HTTP Proxy configuration option

gantrior:
Adding `otel.http_proxy_url` configuration option to values.yaml which will add `HTTP_PROXY` environment variable to every single OTEL collector deployed as part of SWO K8s Collector


pstranak-sw:
Closing. The `HTTP_PROXY` environment variable is not used by gRPC exporters so it would have no effect.

PR #684, opened 2024-08-19
NH-87703 Scrape Istio ingress type

jakub-racek-swi:
agent part of https://swicloud.atlassian.net/browse/NH-87703


pstranak-sw:
You need to update also the list of monitored object types: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/684/files#diff-cea4585ccba473961e2c8a1c18c77d49ca16eaa78a164a00f502ab42a635ce31R226-R237

deploy/helm/events-collector-config.yaml
```diff
@@ -121,7 +121,8 @@ processors:
             - set(attributes["k8s.persistentvolume.name"], body["metadata"]["name"]) where body["kind"] == "PersistentVolume"
             - set(attributes["k8s.persistentvolumeclaim.name"], body["metadata"]["name"]) where body["kind"] == "PersistentVolumeClaim"
             - set(attributes["k8s.namespace.name"], body["metadata"]["namespace"])
-            - set(attributes["k8s.ingress.name"], body["metadata"]["name"]) where body["kind"] == "Ingress"
+            - set(attributes["k8s.ingress.name"], body["metadata"]["name"]) where body["kind"] == "Ingress" or body["kind"] == "VirtualService"
+            - set(attributes["k8s.ingress.type"], body["kind"]) where body["kind"] == "Ingress" or body["kind"] == "VirtualService"
```
pstranak-sw, line 125: `k8s.ingress.type` is not an official OTEL k8s attribute. And there is a real chance that if it is added in the future, its meaning may be different from ours. We should call it `sw.k8s.ingress.type`.

jakub-racek-swi: Sure, if it makes the attribute more futureproof 👍

deploy/helm/events-collector-config.yaml
```diff
@@ -234,6 +235,7 @@ receivers:
   "persistentvolumes"
   "persistentvolumeclaims"
   "ingresses"
+  "virtualservices"
```
pstranak-sw, line 238: Hmm, judging by the source code of the `k8sobjects` receiver, we can list here only objects that we know for sure have definitions in the cluster.
That is OK for the standard resource types, but it is a problem for CRDs, like istio's VirtualService`.

jakub-racek-swi: I've just added an additional conditional check, that looks for the istio API definition.
`{{- if .Capabilities.APIVersions.Has "networking.istio.io/v1alpha3" -}}`
I have tested that this works with/without istio installed, but it ignores the case where the user has istio api but doesn't know virtualservices. I have not found a way to check for a certain CRD without crashing, so this was the best way I found.

jakub-racek-swi: `k8sobjects` gets its object data straight from kubernetes server api, so it should theoretically know everything I can search for with kubectl, i think.

jakub-racek-swi: The issu is that the function only works with a specific api version, and doesn't support wildcards, so we'd have to maintain it with new istio API versions..

pstranak-sw: There are two disadvantages of the `.Capabilities.APIVersions.Has` approach:
* It's not compatible with EKS addons. The EKS addon will never monitor `VirtualService` this way.
* The check is only applied during installation or update of the Helm chart. If Istio is added to the cluster after `k8s collector` is deployed, it won't notice it.

jakub-racek-swi: Can you expand on the issue of the EKS addon?

The second point could theoretically be solved with having some kind of Mutating WebHook or Kubernetes Operator that listens to changes in the cluster, though I don't have experience with using it, and it sounds like too much for this issue?
We could also just notify the user to update the chart if we find istio metrics in a cluster that isn't configured to scrape `VirtualService`, or similarly introduce a flag to values for this, which again, requires user action.

I'm currently not aware of other approaches, sadly.

pstranak-sw (approved):

pstranak-sw (approved):

PR #690, opened 2024-08-26
Use `container-parser` operator in `filelogreceiver` instead of custom configuration

pstranak-sw:
OTEL Collector [v0.101.0](https://github.com/open-telemetry/opentelemetry-collector-contrib/releases/tag/v0.101.0) added a new `container` parser to the `filelog` receiver.
The new parser should provide a unified way for parsing container logs and metadata from them. Plus, it should be more efficient than the combination of other parsers that is used usually.

This PR replaces the old combination of parsers with the new one.

Notes:
* The new parser has some breaking changes (e.g. moving stuff from `attributes` to `resource_attributes`). So, in situations, where customers depend on the old behavior, we still deploy the original configuration. When customers use the filtering syntax (which is a breaking change by itself), or no custom filters at all, we use the new parser.
  The behavior in SWO UI should not be affected. Except for the next point.
* The log records now contain an attribute called `log.iostream`.  It is a replacement for `stream`, compatible with OTEL SemConv. No one should be using that attribute as it is not mentioned in any of our documents. So, I did not try to undo the change in our configuration. But if we are not comfortable with this breaking change, we can rename it back to the old name.
* The new parser does not fully support Windows path, so there is a workaround for that.
* The new `container` parser by default adds a new `time` attribute. Our old implementation was not sending that attribute and the plan for the new parser is also to not send it. So, this PR enables the FF `filelog.container.removeOriginalTimeField` to get that behavior now.

### DIFF
--- a/deploy/helm/CHANGELOG.md
+++ b/deploy/helm/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## Unreleased
 
+### Changed
+
+- Switched to using `container` parser in the `filelog` receiver for parsing container logs and their metadata
+
 ## [4.0.0-alpha.9] - 2024-08-23
 
 ### Changed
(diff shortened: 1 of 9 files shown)


gantrior (approved):
nice

PR #711, opened 2024-09-13
NH-81349 Add node selector to autoupdate cronjob

jakub-racek-swi:
Jira: https://swicloud.atlassian.net/browse/NH-81349

### DIFF
--- a/deploy/helm/templates/autoupdate/autoupdate-cronjob.yaml
+++ b/deploy/helm/templates/autoupdate/autoupdate-cronjob.yaml
@@ -17,6 +17,8 @@ spec:
           imagePullSecrets:
             {{- toYaml .Values.imagePullSecrets | nindent 12 }}
           {{- end }}
+          nodeSelector:
+            kubernetes.io/os: linux
           containers:
           - name: helm-upgrade
             image: "{{ include "common.image" (tuple . .Values.autoupdate (tuple "image" "autoupdate")) }}"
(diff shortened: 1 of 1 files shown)


deploy/helm/templates/autoupdate/autoupdate-cronjob.yaml
```diff
@@ -17,6 +17,8 @@ spec:
           imagePullSecrets:
             {{- toYaml .Values.imagePullSecrets | nindent 12 }}
           {{- end }}
+          nodeSelector:
+            kubernetes.io/os: linux
```
pstranak-sw, line 21: Do we need to include also architecture requirements, like we do for other workloads? For example:
```suggestion
          affinity:
            nodeAffinity:
              requiredDuringSchedulingIgnoredDuringExecution:
                nodeSelectorTerms:
                - matchExpressions:
                  - key: kubernetes.io/arch
                    operator: In
                    values:
                    - amd64
                    - arm64
                  - key: kubernetes.io/os
                    operator: In
                    values:
                    - linux
```

pstranak-sw (approved):

deploy/helm/templates/autoupdate/autoupdate-cronjob.yaml
jakub-racek-swi: I'm not sure if it is needed necessarily, but it could save us from future problems

etichy (approved):

PR #716, opened 2024-09-16
Make `cluster.uid` optional

pstranak-sw:
Make `cluster.uid` optional. Use `cluster.name` if `cluster.uid` is not provided. Make sure it's safe to use in a file path. And add a values check for obviously incorrect `cluster.name` and `cluster.uid` values.

There is a change in behavior of how log checkpoint folders are named. And depending on the provided `cluster.uid` value, the Fargate config can also generate a different log group name.

### DIFF
--- a/deploy/helm/README.md
+++ b/deploy/helm/README.md
@@ -10,7 +10,6 @@
 
 Walk through `Add a Kubernetes cluster` in [SolarWinds Observability](https://documentation.solarwinds.com/en/success_center/observability/default.htm#cshid=gh-k8s-collector)
 
-
 ## Configuration
 
 The [Helm chart](Chart.yaml) that you are about to deploy to your cluster has various configuration options. The full list, including the default settings, is available in [values.yaml](values.yaml).
@@ -33,6 +32,10 @@ In order to deploy the Helm chart, you need to prepare:
       name: <cluster-display-name>
       uid: <unique-cluster-identifier>
   ```
+
+#### Version 4.1.0 and newer
+
+Starting with version 4.1.0 setting `cluster.uid` is optional. If not provided it defaults to value of `cluster.name`.
 
 ### Metrics
 
--- a/deploy/helm/templates/_helpers.tpl
+++ b/deploy/helm/templates/_helpers.tpl
@@ -43,6 +43,15 @@ Usages:
 {{- end -}}
 
 {{/*
+Get cluster UID based on name and uid provided in .Values.cluster.
+Usage:
+  {{ include "common.cluster-uid" . }}
+*/}}
+{{- define "common.cluster-uid" -}}
+{{ default .Values.cluster.name .Values.cluster.uid }}
+{{- end -}}
+
+{{/*
 Common pod labels - those labels are included on every pod in the chart
 */}}
 {{- define "common.pod-labels" -}}
--- a/deploy/helm/templates/logs-fargate-config-map.yaml
+++ b/deploy/helm/templates/logs-fargate-config-map.yaml
@@ -1,7 +1,10 @@
 {{- if and .Values.aws_fargate.enabled .Values.aws_fargate.logs.enabled }}
 {{- $pattern := "[^a-zA-Z0-9\\.\\-_/#]" -}}
 {{- $replacement := "" -}}
-{{- $name := regexReplaceAll $pattern .Values.cluster.name $replacement | default .Values.cluster.uid | lower | trunc 512 -}}
+{{/* Set the log group name either to a sanitized cluster name, or to a cluster uid or to a hash of a cluster name */}}
+{{- $sanitizedName := regexReplaceAll $pattern .Values.cluster.name $replacement -}}
+{{- $sanitizedUid := regexReplaceAll $pattern (default "" .Values.cluster.uid) $replacement -}}
+{{- $name := default (default (sha256sum .Values.cluster.name) $sanitizedUid) $sanitizedName | lower | trunc 512 -}}
 apiVersion: v1
 kind: ConfigMap
 metadata:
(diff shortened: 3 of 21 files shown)


deploy/helm/templates/_helpers.tpl
```diff
@@ -42,6 +42,15 @@ Usages:
 {{- printf "%s%s" $result $suffix -}}
 {{- end -}}
 
+{{/*
+Get cluster UID based on name and uid provided in .Values.cluster.
+Usage:
+  {{ include "common.cluster-uid" . }}
+*/}}
+{{- define "common.cluster-uid" -}}
```
pstranak-sw, line 50: I put the UID logic into this template to make it easier to change it in the future. But right now, with a simple "default", it's not that much beneficial.

deploy/helm/templates/logs-fargate-config-map.yaml
```diff
@@ -1,7 +1,10 @@
 {{- if and .Values.aws_fargate.enabled .Values.aws_fargate.logs.enabled }}
 {{- $pattern := "[^a-zA-Z0-9\\.\\-_/#]" -}}
 {{- $replacement := "" -}}
-{{- $name := regexReplaceAll $pattern .Values.cluster.name $replacement | default .Values.cluster.uid | lower | trunc 512 -}}
+{{/* Set the log group name either to a sanitized cluster name, or to a cluster uid or to a hash of a cluster name */}}
+{{- $sanitizedName := regexReplaceAll $pattern .Values.cluster.name $replacement -}}
+{{- $sanitizedUid := regexReplaceAll $pattern (default "" .Values.cluster.uid) $replacement -}}
+{{- $name := default (default (sha256sum .Values.cluster.name) $sanitizedUid) $sanitizedName | lower | trunc 512 -}}
```
pstranak-sw, line 7: The `sha256sum .Values.cluster.name` was added as a new last fallback, because both `$sanitizedName` and `$sanitizedUid` could contain the same string.

deploy/helm/templates/metrics-deployment.yaml
```diff
@@ -82,8 +82,8 @@ spec:
           command: ['/bin/grpcurl', '-expand-headers',
                     '-proto', 'opentelemetry/proto/collector/logs/v1/logs_service.proto',
                     '-H', 'Authorization: Bearer ${SOLARWINDS_API_TOKEN}',
-                    '-d', '{"resourceLogs":[{"resource":{"attributes":[{"key":"sw.k8s.cluster.uid","value":{"stringValue": "{{ .Values.cluster.uid }}"}}]},"scope_logs":{"log_records":[{"severityText":"INFO","body":{"stringValue":"otel-endpoint-check successful"}}]}}]}',
-                    '{{ .Values.otel.endpoint }}', 'opentelemetry.proto.collector.logs.v1.LogsService.Export']
+                    '-d', '{"resourceLogs":[{"resource":{"attributes":[{"key":"sw.k8s.cluster.uid","value":{"stringValue": "${CLUSTER_UID}"}}]},"scope_logs":{"log_records":[{"severityText":"INFO","body":{"stringValue":"otel-endpoint-check successful"}}]}}]}',
+                    '$(OTEL_ENVOY_ADDRESS)', 'opentelemetry.proto.collector.logs.v1.LogsService.Export']
```
pstranak-sw, line 86: It's a Kubernetes recognized environment variable (`$(...)`), not a shell environment variable (`${...}`). `grpcurl` does not support shell variables for this parameter.

deploy/helm/templates/metrics-deployment.yaml
```diff
@@ -108,6 +108,9 @@ spec:
                   name: {{ template "common.secret" . }}
                   key: SOLARWINDS_API_TOKEN
                   optional: true
+          envFrom:
+            - configMapRef:
+                name: {{ include "common.fullname" (tuple . "-common-env") }}
```
pstranak-sw, line 113: Use the environment variables for `CLUSTER_UID` and `OTEL_ENVOY_ADDRESS`, like we do in all other containers.

deploy/helm/templates/node-collector-daemon-set.yaml
```diff
@@ -168,7 +168,7 @@ spec:
             path: /var/log/journal
         - name: logcheckpoints
           hostPath:
-            path: {{ printf "%s/%s" .Values.otel.logs.filestorage.directory .Values.cluster.uid }}
+            path: {{ printf "%s/%s" .Values.otel.logs.filestorage.directory (sha256sum (include "common.cluster-uid" .)) }}
```
pstranak-sw, line 171: Using `sha256sum` instead of the UID directly, because it could contain problematic characters.
That was true even before this PR, but now, it's much more likely to happen, since cluster names are more likely to contain white spaces, etc.
Is changing the name of the directory going to cause any issues for log collection after a collector's upgrade?

deploy/helm/templates/swo-agent-statefulset.yaml
```diff
@@ -50,7 +50,10 @@ spec:
           imagePullPolicy: {{ .Values.swoagent.image.pullPolicy }}
           env:
             - name: UAMS_CLIENT_ID_OVERRIDE_SOURCE_NAME
-              value: {{ .Values.cluster.uid }}
+              valueFrom:
+                configMapKeyRef:
+                  name: {{ include "common.fullname" (tuple . "-common-env") }}
+                  key: CLUSTER_UID
```
pstranak-sw, line 56: Use the environment variable for CLUSTER_UID, like we do in all other containers.

deploy/helm/tests/logs-fargate-config-map_test.yaml
```diff
@@ -2,6 +2,8 @@
 suite: Test for logs-fargate-config-map
 templates:
   - logs-fargate-config-map.yaml
+chart:
+  version: 1.0.0
```
pstranak-sw, line 6: Added explicit chart version so that the generated `sw.k8s.agent.manifest.version` in test snapshots does not change with each new release of the chart.

deploy/helm/README.md
```diff
@@ -34,6 +33,10 @@ In order to deploy the Helm chart, you need to prepare:
       uid: <unique-cluster-identifier>
```
jaroslav-fedor-swi, line 33:
```suggestion
```

pstranak-sw: See the new sentence just below.

gantrior (approved):

PR #725, opened 2024-09-18
Refactor release notes

jaroslav-fedor-swi:


.github/release.yml
```diff
@@ -0,0 +1,13 @@
+changelog:
+  exclude:
+    labels:
+      - skip-changelog
```
gantrior, line 4: This is what is used in dependabot (still one occurance): `Skip Changelog`

gantrior (approved):

PR #732, opened 2024-09-20
Adds support for RedHat OpenShift clusters

etichy:
Our collector should be able to run on openshift cluster. 

Limitations: 
- ebpf is not running correctly - https://github.com/open-telemetry/opentelemetry-network/issues/254

### DIFF
--- a/deploy/helm/templates/node-collector-daemon-set.yaml
+++ b/deploy/helm/templates/node-collector-daemon-set.yaml
@@ -31,7 +31,9 @@ spec:
     spec:
       terminationGracePeriodSeconds: {{ .Values.otel.node_collector.terminationGracePeriodSeconds }}
       serviceAccountName: {{ include "common.fullname" . }}
+      {{- if .Values.openshift.enabled }}
       hostNetwork: true
+      {{- end }}
       securityContext:
         fsGroup: 0
         runAsUser: 0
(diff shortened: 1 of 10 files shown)


deploy/helm/templates/node-collector-daemon-set.yaml
```diff
@@ -31,11 +31,20 @@ spec:
     spec:
       terminationGracePeriodSeconds: {{ .Values.otel.node_collector.terminationGracePeriodSeconds }}
       serviceAccountName: {{ include "common.fullname" . }}
+      hostNetwork: true
```
jaroslav-fedor-swi, line 34: Do we need to be on host metric everywhere or only on openshift?

deploy/helm/templates/openshift/openshift-role-binding.yaml
```diff
@@ -0,0 +1,20 @@
+{{- if .Values.openshift.enabled }}
+apiVersion: rbac.authorization.k8s.io/v1
+kind: RoleBinding
+metadata:
+  name: {{ include "common.fullname" (tuple . "-openshift-role-binding") }}
```
jaroslav-fedor-swi, line 5: Shoud we name that this RB is for kube-state-metrics?

etichy: it has collector name in prefix, right?
image

deploy/helm/templates/node-collector-daemon-set.yaml
etichy: this is due to INFRA nodes, we can make it conditional only for openshift.

gantrior: Why do we need to mount node collector to host network interface? Even in openShift?

etichy: hostNetwork option solves issues with access to journal logs and wr access to checkpoints folder. We decided that as first draft we go out with this not optimal solution and in operator we will deal with it properly.

gantrior (approved):

PR #759, opened 2024-10-14
Use `kind` for autoupdate test

pstranak-sw:
The used `k3s` stopped working, likely because the build machine switched to a newer Ubuntu release. And since we switched to `kind` in other places anyway, let's replace it in this test, too.


gantrior (approved):

pstranak-sw:
Closing in favor of PR #761

PR #780, opened 2024-11-06
NH-94277: fixing scrape of etcd and controller manager with autodiscovery disabled

gantrior:
`{{- if .Values.otel.metrics.autodiscovery.prometheusEndpoints.enabled}}` should end before etc/controller manager

### DIFF
--- a/deploy/helm/Chart.yaml
+++ b/deploy/helm/Chart.yaml
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: swo-k8s-collector
-version: 4.2.0-alpha.8
+version: 4.2.0-alpha.9
 appVersion: 0.11.7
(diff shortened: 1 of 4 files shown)


pstranak-sw:
Do we still need `prometheus/discovery/https`? And why do we use service account cert in it?

gantrior:
> Do we still need `prometheus/discovery/https`? And why do we use service account cert in it?

Yes in case of `prometheus.io/scheme` of the application metrics is `https`, we try to access it using kubernetes TLS

juraj-paska (approved):

pstranak-sw:
> Yes in case of prometheus.io/scheme of the application metrics is https, we try to access it using kubernetes TLS
I know, but that is an interesting assumption.

PR #787, opened 2024-11-11
NH-90472: Adding OTEL Operator as subchart

gantrior:
* Adding OTEL Operator as chart dependency when `operator.enabled=true` is set.
* Adding Cert Manager as chart dependency when `certmanager.enabled=true` is set.
* `Instrumentation` object is not yet part of the chart, because it would limit instrumentation scope to the namespace where SWO K8s Collector is installed. To explain this: 
    * We could create this kind of resource:
```
{{- if .Values.operator.enabled }}
apiVersion: opentelemetry.io/v1alpha1
kind: Instrumentation
metadata:
  name: swo-instrumentation
  annotations:
    "helm.sh/hook": post-install,post-upgrade
    "helm.sh/hook-weight": "1"
spec:
  java:
    image: ghcr.io/solarwinds/autoinstrumentation-java:2.9.0
    env:
      - name: SW_APM_DEBUG_LEVEL
        value: "trace"
      - name: SW_APM_SERVICE_KEY
        valueFrom:
          secretKeyRef:
            key: SOLARWINDS_API_TOKEN
      - name: SW_APM_COLLECTOR
        value: {{ .Values.otel.endpoint | replace "otel.collector." "apm.collector."  }}
{{- end }}
```

The issue is that `SW_APM_SERVICE_KEY` is injected to target app as is, therefore if it is in different namespace than SWO K8s Collector it wouldn't work. So customer needs to have control over `Instrumentation` objects and secrets.
* Documentated installation steps. 
* Releasing this as 4.3.0-alpha.1


deploy/helm/values.schema.json
```diff
@@ -1254,6 +1254,36 @@
...
+            "description": "If enabled, deploys otel-operator to the cluster.",
+            "type": "object",
+            "properties": {
+                "enabled": {
+                    "type": "boolean"
+                },
+                "additionalProperties": true
+            }
+        },
+        "opentelemetry-operator": {
```
pstranak-sw, line 1268: Why do we need the "legacy" fields in the schema?

deploy/helm/values.yaml
```diff
@@ -785,4 +785,29 @@ ebpfNetworkMonitoring:
...
+      repository: ghcr.io/open-telemetry/opentelemetry-operator/opentelemetry-operator
+      tag: "0.107.0"
+    collectorImage:
+      # it is not yet supported to run collector instrumented by operator
+      # filling this to satisfy subchart requirement
+      repository: "solarwinds/swi-opentelemetry-collector"
+  admissionWebhooks:
+    certManager:
+      enabled: true
+      # ensure that Certificate and Issuer resource are created after CRD's of all dependencies
```
pstranak-sw, line 803: Doesn't Helm create CRDs of a chart and all its dependencies before everything else?

pstranak-sw:
Why do we need an additional dependency of `cert-manager`? Can't we just use `operator.admissionWebhooks.autoGenerateCert.enabled=true`?

deploy/helm/values.yaml
pstranak-sw: No, because the dependencies create CRDs through templates, not using Helm's support for CRDs.

deploy/helm/values.schema.json
pstranak-sw: This is supposed to be a workaround for older versions of the Helm tool. We should check if it actually works.

pstranak-sw:
> Why do we need an additional dependency of `cert-manager`? Can't we just use `operator.admissionWebhooks.autoGenerateCert.enabled=true`?

We could. But there is no autorenewal. The issue is that customers would have to install/upgrade the `k8s collector` in their cluster at least once per year.

pstranak-sw (approved):

deploy/helm/values.schema.json
gantrior: I verified that in CI which uses Helm 3.8.0, and it deployed all certmanager and operator pods correctly (https://github.com/solarwinds/swi-k8s-opentelemetry-collector/actions/runs/11818302623/job/32925550157#step:12:8), so `values.yaml` is respected correctly in older Helm:

pstranak-sw (approved):

pstranak-sw (approved):

PR #788, opened 2024-11-12
Start collecting manifest Config Maps

mpecenka:
collecting manifests for configmaps


jakub-racek-swi (dismissed):

deploy/helm/Chart.yaml
```diff
@@ -1,6 +1,6 @@
 apiVersion: v2
 name: swo-k8s-collector
-version: 4.2.0-alpha.11
+version: 4.2.0-alpha.12
```
gantrior, line 3: I wouldn't include this into 4.2.0, this should be 4.3.0 material. We wanted to close 4.2.0 as soon as we can

gantrior (changes requested):

deploy/helm/Chart.yaml
mpecenka: changed to 4.3.0-alpha.1

gantrior (dismissed):

pstranak-sw (approved):

PR #807, opened 2024-11-26
NH-70086 add init container for OTEL connection to node and event collectors

jakub-racek-swi:
link: https://swicloud.atlassian.net/browse/NH-70086

modified values.yaml and moved 'otel.metrics.swi_endpoint_check' to 'otel.swi_endpoint_check' to be used for all the init containers


deploy/helm/values.schema.json
```diff
@@ -1141,6 +1141,15 @@
                         }
                     },
                     "additionalProperties": false
+                },
+                "swi_endpoint_check": {
+                    "type": "object",
+                    "properties": {
+                        "enabled": {
```
pstranak-sw, line 1148: Please add some description to this property.

pstranak-sw:
Is there a reason for not adding this also for the node collector for Windows?

pstranak-sw:
Please make sure there are unit tests for the `otel.swi_endpoint_check.enabled` and `otel.metrics.swi_endpoint_check` setting for all the Deployments/DaemonSets, not just the metric collector.

deploy/helm/templates/events-collector-deployment.yaml
```diff
@@ -63,6 +63,29 @@ spec:
       imagePullSecrets:
         {{- toYaml .Values.imagePullSecrets | nindent 8 }}
       {{- end }}
+      {{- if and .Values.otel.swi_endpoint_check.enabled .Values.otel.metrics.swi_endpoint_check }}
```
gantrior, line 66: If the old property gets deprecated we don't want to check for both, we want to use the new value if it's set, and only use the old value if the new one isn't provided.

gantrior (commented):
If the old value is deprecated we need to communicate it and we should remove it from `values.yaml` (but continue support it in templates). 

I recommend to use following condition (haven't tested): 
```
{{- if .Values.otel.swi_endpoint_check.enabled | default .Values.otel.metrics.swi_endpoint_check | default true }}
```
and we should also put to NOTES.txt deprecation warning, something like this:
```
{{- if and (not (hasKey .Values.otel "swi_endpoint_check")) (hasKey .Values.otel.metrics "swi_endpoint_check") }}
  {{- $warning := printf "Value 'otel.metrics.swi_endpoint_check' is deprecated and will be removed in a future release. Please use 'otel.swi_endpoint_check.enabled' instead." }}
  {{- warn $warning }}
{{- end }}
```

pstranak-sw:
> If the old value is deprecated we need to communicate it and we should remove it from `values.yaml` (but continue support it in templates).
> 
> ...

Good point. The condition check was adjusted, though using a different logic.

deploy/helm/templates/NOTES.txt
```diff
@@ -7,7 +7,7 @@ WARNING: Prometheus is no longer included in this chart. To scrape custom metric
 
 {{- if and .Values.otel.metrics.extra_scrape_metrics (and .Values.otel.metrics.autodiscovery.prometheusEndpoints.enabled (not .Values.otel.metrics.force_extra_scrape_metrics)) -}}
 WARNING: You have enabled autodiscovery of prometheus endpoints, so `extra_scrape_metrics` is ignored. If you are sure that those metrics won't be covered by autodiscovery set `otel.metrics.force_extra_scrape_metrics` to `true`.
-
+{{- println -}}
```
pstranak-sw, line 10: Added to fix NOTES.txt unit tests.

deploy/helm/templates/_helpers.tpl
```diff
@@ -376,3 +376,13 @@ YAML with the filter.
...
 
 {{- end -}}
+
+{{/*
+Check whether the SWI endpoint check is enabled
+
+Usage:
+{{ isSwiEndpointCheckEnabled . }}
+*/}}
+{{- define "isSwiEndpointCheckEnabled" -}}
```
pstranak-sw, line 386: I don't really like templates, but the condition was too complex and in too many places not to create a template for it.

deploy/helm/templates/metrics-deployment.yaml
```diff
@@ -60,11 +60,11 @@ spec:
                 values:
                 - linux
       {{- end }}
-      {{- if or (and .Values.otel.metrics.prometheus_check .Values.otel.metrics.prometheus.url) .Values.otel.metrics.swi_endpoint_check }}
       {{- if .Values.imagePullSecrets }}
       imagePullSecrets:
```
pstranak-sw, line 64: I'm not sure why `imagePullSecrets` was under the init containers condition and not separate, as it is in other deployments.

src/cmd/connectioncheck/connection_check.go
```diff
@@ -51,6 +51,7 @@ func sendTestMessage(endpoint, apiToken, clusterUid string, insecure bool) {
 	record.SetTimestamp(time.Now())
 
 	logger.Emit(ctx, record)
+	log.Print("Connection check was successful")
```
pstranak-sw, line 54: Added so that the init container shows some success message instead of just ending. However, this change will actually be applied only after we release a new image.

PR #812, opened 2024-11-28
Autodiscovery for Fargate pods

mpecenka:
When aws_fargate.enabled is true, spawn deployment that will do discovery on Fargate pods.

### DIFF
--- a/deploy/helm/metrics-discovery-config.yaml
+++ b/deploy/helm/metrics-discovery-config.yaml
@@ -16,8 +16,8 @@ exporters:
       queue_size: {{ .Values.aws_fargate.metrics.autodiscovery.sending_queue.queue_size }}
 {{- if .Values.aws_fargate.metrics.autodiscovery.sending_queue.offload_to_disk }}
       storage: file_storage/sending_queue
-    timeout: {{ .Values.aws_fargate.metrics.autodiscovery.timeout }}
 {{- end }}
+    timeout: {{ .Values.aws_fargate.metrics.autodiscovery.timeout }}
 extensions:
 {{- if .Values.aws_fargate.metrics.autodiscovery.sending_queue.offload_to_disk }}
   file_storage/sending_queue:
--- a/deploy/helm/metrics-discovery-config.yaml
+++ b/deploy/helm/metrics-discovery-config.yaml
@@ -25,6 +25,10 @@ extensions:
 {{- end }}
   health_check:
     endpoint: 0.0.0.0:13133
+  k8s_observer:
+    auth_type: serviceAccount
+    node: ${NODE_NAME}
+    observe_pods: true    
 
 processors:
   swk8sattributes:
@@ -174,14 +178,6 @@ processors:
         action: update
         new_name: k8s.$${1}$${2}
   
-  metricstransform/rename-otel:
-    transforms:
-      # add `k8s.` prefix to all metrics
-      - include: ^(.*)$$
-        match_type: regexp
-        action: update
-        new_name: k8s.$${1}
-
   # Transformations done on all metrics before any grouping
   metricstransform/preprocessing:
     transforms:
@@ -772,26 +768,6 @@ processors:
 
   {{- include "common-config.attributes-remove-prometheus-attributes" . | nindent 2 }}
 
-  resource/otlp-metrics:
-    attributes:
-      # Collector and Manifest version
-      - key: sw.k8s.agent.manifest.version
-        value: ${MANIFEST_VERSION}
-        action: insert
-
-      - key: sw.k8s.agent.app.version
-        value: ${APP_VERSION}
-        action: insert
-
-      # Cluster
-      - key: sw.k8s.cluster.uid
-        value: ${CLUSTER_UID}
-        action: insert
-
-      - key: k8s.cluster.name
-        value: ${CLUSTER_NAME}
-        action: insert
-
   batch:
 {{ toYaml .Values.otel.metrics.batch | indent 4 }}
   filter/kube-state-metrics:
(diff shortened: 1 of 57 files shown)


deploy/helm/templates/metrics-discovery-deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+{{- if .Values.aws_fargate.discovery.telemetry.metrics.enabled }}
+            - name: http
+              containerPort: {{ (split ":" .Values.aws_fargate.discovery.telemetry.metrics.address)._1 }}
+              protocol: TCP
+{{- end}}
+          envFrom:
+            - configMapRef:
+                name: {{ include "common.fullname" (tuple . "-common-env") }}
+            - configMapRef:
+                name: {{ include "common.fullname" (tuple . "-metrics-env-config") }}
```
pstranak-sw, line 146: Should this be here?

deploy/helm/templates/metrics-discovery-deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+            seccompProfile:
+              type: RuntimeDefault
+            capabilities:
+              drop:
+              - ALL
+
+          image: "{{ include "common.image" (tuple . .Values.otel "image" nil .Chart.AppVersion) }}"
+          imagePullPolicy: {{ .Values.otel.image.pullPolicy }}
+          env:
+            - name: MY_POD_IP
```
pstranak-sw, line 118: What do we use this for?

deploy/helm/templates/metrics-discovery-deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+        {{- if and .Values.otel.metrics.prometheus_check .Values.otel.metrics.prometheus.url }}
+        - name: prometheus-check
+          image: "{{ include "common.image" (tuple . .Values.otel.init_images "busy_box") }}"
+          imagePullPolicy: {{ .Values.otel.init_images.busy_box.pullPolicy }}
+          command: ['sh', '-c', 'until $(wget --spider -nv $PROMETHEUS_URL/federate?); do echo waiting on prometheus; sleep 1; done && echo "Prometheus is available"']
+          envFrom:
+            - configMapRef:
+                name: {{ include "common.fullname" (tuple . "-metrics-env-config") }}
+        {{- end }}
+        {{- if include "isSwiEndpointCheckEnabled" . }}
```
pstranak-sw, line 77: We don't use support Prometheus server for discovery. And the `$PROMETHEUS_URL` variable should not even be initialized. I think this does not belong here.

deploy/helm/templates/metrics-discovery-deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+  name: {{ include "common.fullname" (tuple . "-metrics-discovery") }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+    app.kubernetes.io/name: swo-k8s-collector
+{{ include "common.labels" . | indent 4 }}
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app: {{ include "common.fullname" (tuple . "-metrics") }}
```
pstranak-sw, line 14: Wrong label.

deploy/helm/templates/metrics-discovery-deployment.yaml
```diff
@@ -0,0 +1,182 @@
...
+                  key: SOLARWINDS_API_TOKEN
+                  optional: true
+            - name: GOMEMLIMIT
+              valueFrom:
+                resourceFieldRef:
+                  resource: limits.memory
+          ports:
+            - name: otlp
+              containerPort: {{ .Values.otel.metrics.otlp_endpoint.port }}
+              protocol: TCP
```
pstranak-sw, line 136: We don't need this endpoint for discovery, right?

deploy/helm/metrics-discovery-config.yaml
```diff
@@ -0,0 +1,274 @@
...
+      processors:
+        - memory_limiter
+        - filter/histograms
+        - swk8sattributes
+{{- if not (empty .Values.otel.metrics.k8s_instrumentation.labels.excludePattern) }}
+        - resource/swk8sattributes_labels_filter
+{{- end }}
+{{- if not (empty .Values.otel.metrics.k8s_instrumentation.annotations.excludePattern) }}
+        - resource/swk8sattributes_annotations_filter
+{{- end }}
```
pstranak-sw, line 256: This is no longer necessary.

pstranak-sw (changes requested):

deploy/helm/values.yaml
```diff
@@ -708,6 +708,83 @@ aws_fargate:
     # NOTE: The FluentBit configuration expects four spaces as indentation within sections
     filters: ""
 
+  # Configuration for Fargate discovery deployment
+  discovery:
```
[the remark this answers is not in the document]
gantrior: maybe use the same path? like:
```
metrics:
  autodiscovery:
     ....
```

pstranak-sw:
@gantrior, do we need to create PodMonitor for the new deployment? Like https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/templates/metrics-pod-monitor.yaml ?

deploy/helm/metrics-discovery-config.yaml
```diff
@@ -0,0 +1,268 @@
...
+
+  batch/metrics:
+{{ toYaml .Values.aws_fargate.metrics.autodiscovery.batch | indent 4 }}
+  
+{{- include "common-config.filter-remove-temporary-metrics" . | nindent 2 }}
+
+  filter/histograms:
+    metrics:
+      metric:
+        - 'type == METRIC_DATA_TYPE_HISTOGRAM and not(name == "k8s.apiserver_request_duration_seconds" or name == "k8s.workqueue_queue_duration_seconds" or name == "k8s.coredns_dns_request_duration_seconds" or name == "k8s.coredns_dns_request_size_bytes" or name == "k8s.coredns_dns_response_size_bytes")'
```
pstranak-sw, line 100: Are the allowed exceptions correct? The list used in autodiscovery is different.

deploy/helm/metrics-discovery-config.yaml
```diff
@@ -0,0 +1,268 @@
...
+{{- range .Values.otel.metrics.autodiscovery.prometheusEndpoints.customTransformations.counterToRate }}
+      - {{ . }}
+{{- end }}
+{{- end }}
+
+{{- if ne .Values.otel.metrics.autodiscovery.prefix "k8s." }}
+  # in case the prefix differs from "k8s." we need to copy the required metrics
+  # so that SWO built-in dashboards works correctly
+{{- $arrayOfRequiredMetrics := list 
+  "etcd_disk_backend_commit_duration_seconds"
```
pstranak-sw, line 127: Are the etcd metrics relevant here?

deploy/helm/metrics-discovery-config.yaml
```diff
@@ -0,0 +1,268 @@
...
+      initial_interval: {{ .Values.aws_fargate.metrics.autodiscovery.retry_on_failure.initial_interval }}
+      max_interval: {{ .Values.aws_fargate.metrics.autodiscovery.retry_on_failure.max_interval }}
+      max_elapsed_time: {{ .Values.aws_fargate.metrics.autodiscovery.retry_on_failure.max_elapsed_time }}
+    sending_queue:
+      enabled: {{ .Values.aws_fargate.metrics.autodiscovery.sending_queue.enabled }}
+      num_consumers: {{ .Values.aws_fargate.metrics.autodiscovery.sending_queue.num_consumers }}
+      queue_size: {{ .Values.aws_fargate.metrics.autodiscovery.sending_queue.queue_size }}
+{{- if .Values.aws_fargate.metrics.autodiscovery.sending_queue.offload_to_disk }}
+      storage: file_storage/sending_queue
+    timeout: {{ .Values.aws_fargate.metrics.autodiscovery.timeout }}
```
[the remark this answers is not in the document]
pstranak-sw: Hmm, fixing it also in other places changes the default behavior of the `k8s collector`. This was introduced in #558.

@gantrior What do you think about including this fix in this PR?

pstranak-sw (approved):

gantrior:
> @gantrior, do we need to create PodMonitor for the new deployment? Like https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/templates/metrics-pod-monitor.yaml ?

No I don't think so. Since we no longer encourage users to have Prometheus in the environment, PodMonitors are deprecated and could be removed from Helm chart.

deploy/helm/metrics-discovery-config.yaml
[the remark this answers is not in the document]
gantrior: @pstranak-sw good catch. To not block the verification of this ticket, as this is not related to fargate and discovery I can apply the fix in followup PR

gantrior (approved):

deploy/helm/metrics-discovery-config.yaml
gantrior: It would be absurd environment if someone with Fargate installs etcd separately, since etcd is managed by AWS. But for sake of consistency I would keep same metric list here

PR #833, opened 2024-12-10
Make storageClassName configurable

etichy:


deploy/helm/values.schema.json
```diff
@@ -250,6 +250,13 @@
                 "type": "string"
             }
         },
+        "storageClassName": {
+            "description": "Kubernetes storage classes. More info: https://kubernetes.io/docs/concepts/storage/storage-classes",
+            "type": "object",
+            "additionalProperties": {
+               "type": "string"
+            }
+        },
```
pstranak-sw, line 259: We add `k8s` prefix to definitions, that are defined by Kubernetes. Also, the definition itself seems wrong. The `storageClassName` is an optional string, not an objet.
```suggestion
        "k8sStorageClassName": {
            "description": "The name of the StorageClass required by the claim. More info: https://kubernetes.io/docs/concepts/storage/persistent-volumes#class",
            "type": [
                "string",
                "null"
            ]
        },
```

deploy/helm/values.schema.json
```diff
@@ -1312,6 +1319,9 @@
                 },
                 "resources": {
                     "$ref": "#/definitions/k8sResourceRequirements"
+                },
+                "storageClassName": {
+                    "$ref": "#/definitions/storageClassName"
```
pstranak-sw, line 1324: related to the other comment:
```suggestion
                    "$ref": "#/definitions/k8sStorageClassName"
```

deploy/helm/values.schema.json
pstranak-sw, line 1: Also add the default (undefined) value to the `values.yaml`. Not strictly necessary, but we add all default values there.
```diff
diff --git a/deploy/helm/values.yaml b/deploy/helm/values.yaml
index ef0b673..421ff19 100644
--- a/deploy/helm/values.yaml
+++ b/deploy/helm/values.yaml
@@ -690,6 +690,7 @@ swoagent:
       cpu: 100m
   nodeSelector: {}
   affinity: {}
+  storageClassName:
 
 aws_fargate:
   # Enable support for AWS EKS Fargate environment
```

pstranak-sw (changes requested):

deploy/helm/values.schema.json
etichy: done, thanks 👍

pstranak-sw (approved):

PR #835, opened 2024-12-11
Update Collector to send Values in config map

turytsia:


gantrior (approved):

deploy/helm/events-collector-config.yaml
```diff
@@ -311,7 +370,7 @@ service:
       receivers:
         - k8s_events
 {{- end }}
-{{- if .Values.otel.manifests.enabled }}
+{{- if and .Values.otel.events.enabled .Values.otel.manifests.enabled }}
```
gantrior, line 373: This condition makes no sense anymore, whole events deployment has no condition and we want this pipeline to run even if both manifests and events are disabled no?

gantrior (commented):
The solution is not ideal, there are now two k8sobjects receivers which will cause that config maps will be watched and polled twice. What we should do is to have single `k8sobjects` receiver, and in case manifests are disabled reconfigure that reciever to watch only configmaps. Is there something what I am missing?

deploy/helm/events-collector-config.yaml
```diff
@@ -183,14 +187,30 @@ processors:
...
+
+            # in watch mode body has `object` attribute with the manifest, in pull mode body is the manifest
+            # flaten body structure to have only raw manifest in body in all the cases
+            - set(body, body["object"]) where IsMap(body) and body["object"] != nil
+
+            - set(attributes["k8s.configmap.name"], body["metadata"]["name"]) where body["kind"] == "ConfigMap"
+
+            - set(attributes["k8s.namespace.name"], body["metadata"]["namespace"])
+
+            - set(attributes["k8s.object.kind"], body["kind"]) where body["kind"] != nil
```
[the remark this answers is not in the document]
turytsia: I will be removing this condition. I prefer to keep it as it was

gantrior (dismissed):

deploy/helm/events-collector-config.yaml
```diff
@@ -324,14 +329,16 @@ service:
         - resource/manifest
         - swk8sattributes
         - transform/cleanup_attributes_for_nonexisting_entities
+{{- if not (and .Values.otel.events.enabled .Values.otel.manifests.enabled) }}
```
pstranak-sw, line 332: This means that in the case of disabled event/manifest collection, we will still collects and process everything, just throw almost everything out at the end. It will work, just not efficiently.

pstranak-sw (approved):

deploy/helm/events-collector-config.yaml
turytsia: I will create a follow up PR for this

PR #842, opened 2024-12-17
Fix rendering of node collector

pstranak-sw:


deploy/helm/templates/node-collector-daemon-set-windows.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if and .Values.otel.logs.enabled .Values.otel.windows.enabled }}
+{{- if and (or .Values.otel.logs.enabled .Values.otel.metrics.enabled) .Values.otel.windows.enabled }}
```
pstranak-sw, line 1: Likely forgotten in #473.

mpecenka (approved):

PR #846, opened 2024-12-19
Require Helm 3.9.0 or newer

pstranak-sw:
* Increase required Helm version from 3.8 to 3.9.
* Explicitly mention it in the documentation.
* Check for the version during chart installation.

Note: Helm 3.9.0 is the first version where the `.Capabilities.HelmVersion.Version` actually works.

### DIFF
--- a/deploy/helm/templates/asserts.yaml
+++ b/deploy/helm/templates/asserts.yaml
@@ -9,3 +9,11 @@
 {{- if mustRegexMatch "^[\\s]+$" .Values.cluster.uid -}}
 {{ fail "If specified, the custom cluster UID should be a valid string." }}
 {{- end -}}
+
+{{- if empty .Capabilities.HelmVersion.Version -}}
+{{ fail "This version of Helm is not supported. Please use 3.9.0 or newer." }}
+{{- end -}}
+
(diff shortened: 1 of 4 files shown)


deploy/helm/templates/asserts.yaml
```diff
@@ -8,4 +8,12 @@
 
 {{- if mustRegexMatch "^[\\s]+$" .Values.cluster.uid -}}
 {{ fail "If specified, the custom cluster UID should be a valid string." }}
+{{- end -}}
+
+{{- if empty .Capabilities.HelmVersion.Version -}}
+{{ fail "This version of Helm is not supported. Please use 3.9.0 or newer." }}
+{{- end -}}
+
+{{- if eq ((semver "3.9.0").Compare (semver .Capabilities.HelmVersion.Version)) 1 -}}
```
pstranak-sw, line 17: Technically speaking, this check is not needed, since right now, it only applies to 3.9.0-rc.1 (everything older is also covered by the check above). But I'd keep it to make the requirement explicitly stated here in the code.

jaroslav-fedor-swi (approved):

PR #851, opened 2024-12-27
Remove temporary labels

turytsia:
## Summary

This PR removes temporary attributes `sw.k8s.persistentvolume.found`, `sw.k8s.persistentvolumeclaim.found` and `sw.k8s.service.found`


pstranak-sw:
What about https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/events-collector-config.yaml#L91 ? Do we need to add this there, too?

pstranak-sw (approved):

PR #853, opened 2024-12-31
Read pod status from manifests

turytsia:
## Summary

This PR adds `sw.k8s.pod.status` from pod's manifests


deploy/helm/events-collector-config.yaml
```diff
@@ -116,6 +116,17 @@ processors:
...
+    error_mode: ignore
+    log_statements:
+      - context: log
+        statements:
+          # status
+          - set(attributes["sw.k8s.pod.status"], body["status"]["phase"]) where body["kind"] == "Pod" and body["status"]["phase"] != nil
+
+          # status reason
+          - set(attributes["sw.k8s.pod.phaseReason"], body["status"]["containerStatuses"][0]["state"]["waiting"]["reason"]) where body["kind"] == "Pod" and body["status"]["containerStatuses"][0]["state"]["waiting"]["reason"] != nil
+          - set(attributes["sw.k8s.pod.phaseReason"], body["status"]["containerStatuses"][0]["state"]["terminated"]["reason"]) where body["kind"] == "Pod" and body["status"]["containerStatuses"][0]["state"]["terminated"]["reason"] != nil
```
etichy, line 129: what if there's multiple containers? the failed is always first?

deploy/helm/events-collector-config.yaml
```diff
@@ -116,6 +116,17 @@ processors:
         statements:
           # convert body to string
           - set(body, body.string)
+  transform/set_status_for_pods:
+    error_mode: ignore
+    log_statements:
+      - context: log
+        statements:
+          # status
+          - set(attributes["sw.k8s.pod.status"], body["status"]["phase"]) where body["kind"] == "Pod" and body["status"]["phase"] != nil
```
pstranak-sw, line 125: This creates the attribute on the log level. We need it to be on the resource level - either put it in resource attributes explicitly or group by it to get it there implicitly.

pstranak-sw, line 125: Is the phase value here identical to the value passed in the metric kube_pod_status_phase? No change in letter casing etc.?

deploy/helm/events-collector-config.yaml
turytsia: I don't think it is possible to sort it or iterate through this somehow. We might need a new processor for that.

pstranak-sw (approved):

PR #856, opened 2025-01-06
Update OTEL Collector to v116

pstranak-sw:


gantrior:
What about also pick next version of the image and upgrading also `AppVersion` in `metadata.go`? So that we won't need to bump that in separate PR before publishing the image to docker image

gantrior (approved):

pstranak-sw:
> What about also pick next version of the image and upgrading also `AppVersion` in `metadata.go`? So that we won't need to bump that in separate PR before publishing the image to docker image

I thought about it but decided not to do it, thus avoiding risking merge conflicts with other people.

PR #857, opened 2025-01-06
Update kube-state-metrics sub-chart to 5.28.0

pstranak-sw:
The new version removes two deprecated metrics `kube_endpoint_address_not_ready` and `kube_endpoint_address_available`.

### DIFF
--- a/tests/deploy/tests/test_metric_manifest.yaml
+++ b/tests/deploy/tests/test_metric_manifest.yaml
@@ -69,6 +69,9 @@ spec:
       - name: test-container
         image: busybox
         command: ["sh", "-c", "sleep infinity"]
+        ports:
+        - containerPort: 80
+          name: http-web-svc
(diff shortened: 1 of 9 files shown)


gantrior (dismissed):

tests/deploy/tests/test_metric_manifest.yaml
```diff
@@ -197,7 +200,7 @@ metadata:
     example.com/annotation: "example-annotation"
 spec:
   selector:
-    app: test-pod
+    app: test-statefulset
```
pstranak-sw, line 203: The `kube_endpoint_address` is generated only when there is an actual endpoint mapped to the service. Fixing the link between `test-statefulset` and `test-service`, since the StatefulSet already pointed to Service, just not the other way around.

gantrior (approved):

PR #859, opened 2025-01-08
Remove pod status from metrics

turytsia:
This PR removes `sw.k8s.pod.status` from `metrics-collector-config.yaml`


pstranak-sw:
What about this? https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/1a7f628e9299bb628e464d88ec6aa4ac588e3cf7/deploy/helm/templates/_common-config.tpl#L434

pstranak-sw (approved):

PR #860, opened 2025-01-09
Show error when autoupdate is enabled with pinned image

hadvand:
When the k8s collector is deployed with autoupdate.enabled, fail in case there’s pinned any image in values.yaml.

Changed _helpers.tpl to validate values.yaml images with saved valid images.


hadvand:
possible improvement - move images to separate yaml and pull them from there

PR #865, opened 2025-01-10
Added comments and a warning for enabled autoupdate

hadvand:
This PR adds a comment to values.yaml to carefully consider enabling autoupdate since it can't work with images pinned by user.
Also added a warning to NOTES in case the autoupdate is enabled.


deploy/helm/templates/NOTES.txt
```diff
@@ -96,3 +96,8 @@ WARNING: Configuration `prometheus-node-exporter` was deprecated. Please remove
 WARNING: Value `otel.metrics.swi_endpoint_check` was deprecated and will be removed in a future release. Please use `otel.swi_endpoint_check.enabled` instead.
 {{- println -}}
 {{- end -}}
+
+{{- if .Values.autoupdate.enabled -}}
```
pstranak-sw, line 100: Please add a test for this in https://github.com/solarwinds/swi-k8s-opentelemetry-collector/blob/master/deploy/helm/tests/notes_test.yaml

PR #866, opened 2025-01-11
Turn timeseries-mock-service into a Helm chart

pstranak-sw:
This solves a `skaffold` issue, where the `timeseries-mock-service` had to start after the main chart. That caused a lot of initialization errors and forced us to disable liveness probes in the main chart in tests and local development.


gantrior (approved):
nice improvement

.github/workflows/buildAndTestHelm.yml
```diff
@@ -196,12 +195,11 @@ jobs:
             --create-namespace \
             --namespace swo-k8s-collector \
             --set cluster.name=test-cluster \
-            --set otel.endpoint=timeseries-mock-service:9082 \
+            --set otel.endpoint=timeseries-mock-service.default.svc.cluster.local:9082 \
```
pstranak-sw, line 198: The previous endpoint was incorrect. But because the endpoint checks were disabled, the collector was silently crashing - it was working just enough to pass this autoupdate test. Which also meant that the "timeseries-mock-service" deployed in one of the previous steps wasn't actually used in this test.
Now that the checks are enabled, the setup had to be corrected.

PR #869, opened 2025-01-14
Added warning if autoupdate is enabled

hadvand:
Added warning comment to autoupdate section in values.yaml
Also added a warning to NOTES in case the autoupdate is enabled.


gantrior:
Can we check if `otel.image.tag` or `otel.windows.image.tag` is filled and show error if `autoupdate.enabled` is set to true in the same time? There is also recommended message in JIRA description by @pstranak-sw :
> FATAL: This solarwinds/swo-k8s-collector was configured to check for updates and apply them automatically.
Please make sure none of its container images are pinned to a specific version - it could cause unexpected issues after an update.

hadvand:
@gantrior This was implemented in [previous PR](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/860).
But after discussion we have decided that it's not worth it to maintain this feature because of some downsides (e.g. updating each image in both _helpers.tpl and values.yaml). And some warnings will be enough, since there is still no way to restrict user from changing image in both values.yaml and _helpers.tpl to bypass validation.

PR #876, opened 2025-01-24
Detect sidecar containers

pstranak-sw:
Add attribute `sw.k8s.container.sidecar` to metric `kube_pod_init_container_info`.

PR #878, opened 2025-01-29
Add container processor to send status events

janatrelakova:
# Summary
Adding processor so k8s collector is able to detect the states of pod's containers. Additionally, adding the information whether container is init or sidecar.

### Timestamp of new Log Record
- :question: timestamp is set as the last condition
  - we could fill the timestamp (timeUnixNano) with the time of the received log, but that would not be the real time when the change happened
  - on the other hand, if we take the timestamp of receiver manifest log record maybe it would be better to be transparent from in which log the new log records originated
### Naming
- :question: naming: `containerprocessor` vs `containersprocessor` vs something else
### Container creation
- :thought_balloon: the merging of `spec` and `status` containers is based on `name`
  - container name should be unique across the namespace afaik, so it should be OK to use it as a common denominator
  -   we need to go through `spec` to identify sidecar containers
- :question: I am not sure how you want to handle `sidecar` containers. Do you want to treat `sidecar` container as `init` container?
- if container's name is not in `spec` part, the container will not be included in the resulting log records
### Additional notes
- I have tried to stay with modules version already used in swk8sprocessor
- there are parts that could be parallelized, but I will probably leave it to follow-up
## Example
<details>
<summary> 💡 Example of modified resource log structure</summary>
The first resourceLog is tied to Pod. The second contains processed information about containers contained in the first resourceLog.

Some workload names are anonymized.
```
{
    "resourceLogs": [
        {
            "resource": {
                "attributes": [
                    {
                        "key": "k8s.pod.name",
                        "value": {
                            "stringValue": "test-cronjob-*"
                        }
                    },
                    {
                        "key": "k8s.namespace.name",
                        "value": {
                            "stringValue": "test-namespace"
                        }
                    },
                    {
                        "key": "k8s.pod.annotations.test-annotation",
                        "value": {
                            "stringValue": "test-value"
                        }
                    },
                    {
                        "key": "k8s.pod.labels.app",
                        "value": {
                            "stringValue": "test-cronjob"
                        }
                    },
                    {
                        "key": "sw.k8s.pod.status",
                        "value": {
                            "stringValue": "Succeeded"
                        }
                    },
                    {
                        "key": "sw.k8s.agent.manifest.version",
                        "value": {
                            "stringValue": "4.4.0-alpha.2"
                        }
                    },
                    {
                        "key": "sw.k8s.agent.app.version",
                        "value": {
                            "stringValue": "0.12.3"
                        }
                    },
                    {
                        "key": "sw.k8s.cluster.uid",
                        "value": {
                            "stringValue": "cluster-uid-*"
                        }
                    },
                    {
                        "key": "k8s.cluster.name",
                        "value": {
                            "stringValue": "cluster name"
                        }
                    },
                    {
                        "key": "sw.k8s.log.type",
                        "value": {
                            "stringValue": "manifest"
                        }
                    },
                    {
                        "key": "k8s.node.name",
                        "value": {
                            "stringValue": "docker-desktop"
                        }
                    },
                    {
                        "key": "k8s.cronjob.name",
                        "value": {
                            "stringValue": "test-cronjob"
                        }
                    },
                    {
                        "key": "k8s.job.name",
                        "value": {
                            "stringValue": "test-cronjob-*"
                        }
                    }
                ]
            },
            "scopeLogs": [
                {
                    "scope": {},
                    "logRecords": [
                        {
                            "observedTimeUnixNano": "1738685653111071505",
                            "body": {
                                "stringValue": "<manifest>"
                            },
                            "attributes": [
                                {
                                    "key": "k8s.object.kind",
                                    "value": {
                                        "stringValue": "Pod"
                                    }
                                },
                                {
                                    "key": "watch.type",
                                    "value": {
                                        "stringValue": "MODIFIED"
                                    }
                                },
                                {
                                    "key": "sw.namespace",
                                    "value": {
                                        "stringValue": "sw.events.inframon.k8s.manifests"
                                    }
                                }
                            ],
                            "traceId": "",
                            "spanId": ""
                        }
                    ]
                }
            ]
        },
        {
            "resource": {
                "attributes": [
                    {
                        "key": "sw.k8s.log.type",
                        "value": {
                            "stringValue": "manifest"
                        }
                    }
                ]
            },
            "scopeLogs": [
                {
                    "scope": {
                        "attributes": [
                            {
                                "key": "otel.entity.event_as_log",
                                "value": {
                                    "boolValue": true
                                }
                            }
                        ]
                    },
                    "logRecords": [
                        {
                            "timeUnixNano": "1738685640000000000",
                            "body": {},
                            "attributes": [
                                {
                                    "key": "otel.entity.event.type",
                                    "value": {
                                        "stringValue": "entity_state"
                                    }
                                },
                                {
                                    "key": "otel.entity.type",
                                    "value": {
                                        "stringValue": "KubernetesContainer"
                                    }
                                },
                                {
                                    "key": "otel.entity.id",
                                    "value": {
                                        "kvlistValue": {
                                            "values": [
                                                {
                                                    "key": "k8s.pod.name",
                                                    "value": {
                                                        "stringValue": "test-cronjob-*"
                                                    }
                                                },
                                                {
                                                    "key": "k8s.namespace.name",
                                                    "value": {
                                                        "stringValue": "test-namespace"
                                                    }
                                                },
                                                {
                                                    "key": "k8s.container.name",
                                                    "value": {
                                                        "stringValue": "test-container"
                                                    }
                                                },
                                                {
                                                    "key": "sw.k8s.cluster.uid",
                                                    "value": {
                                                        "stringValue": "cluster-uid-*"
                                                    }
                                                }
                                            ]
                                        }
                                    }
                                },
                                {
                                    "key": "otel.entity.attributes",
                                    "value": {
                                        "kvlistValue": {
                                            "values": [
                                                {
                                                    "key": "container.id",
                                                    "value": {
                                                        "stringValue": "docker://a2c0a0*"
                                                    }
                                                },
                                                {
                                                    "key": "sw.k8s.container.status",
                                                    "value": {
                                                        "stringValue": "terminated"
                                                    }
                                                },
                                                {
                                                    "key": "sw.k8s.container.init",
                                                    "value": {
                                                        "boolValue": false
                                                    }
                                                },
                                                {
                                                    "key": "sw.k8s.container.sidecar",
                                                    "value": {
                                                        "boolValue": false
                                                    }
                                                }
                                            ]
                                        }
                                    }
                                }
                            ],
                            "traceId": "",
                            "spanId": ""
                        }
                    ]
                }
            ]
        }
    ]
}
```
</details>

## TODO
- remove ignoring of resource detection
- devtest


src/processor/containerprocessor/log_constructor.go
```diff
@@ -0,0 +1,77 @@
...
+	// Attributes for OTel entity events identification
+	otelEntityEventAsLog = "otel.entity.event_as_log"
+	otelEntityEventType  = "otel.entity.event.type"
+	swEntityType         = "otel.entity.type"
+
+	// Attributes for telemetry mapping
+	otelEntityId     = "otel.entity.id"
+	k8sContainerName = "k8s.container.name"
+	k8sNamespaceName = "k8s.namespace.name"
+	k8sPodName       = "k8s.pod.name"
```
gantrior, line 21: There is `semcov` package in opentelemetry which already includes many of these, so you don't need to define your constants. See https://pkg.go.dev/go.opentelemetry.io/collector/semconv@v0.118.0/v1.27.0 

So you simply import `semconv "go.opentelemetry.io/collector/semconv/v1.27.0"` and then you can use `semconv.AttributeK8SNamespaceName` etc

src/processor/containerprocessor/log_constructor.go
```diff
@@ -0,0 +1,77 @@
...
+	k8sContainerInit     = "sw.k8s.container.init"
+	k8sContainerSidecar  = "sw.k8s.container.sidecar"
+	k8sContainerId       = "container.id"
+)
+
+func NewContainerResourceLogs() plog.ResourceLogs {
+	rl := plog.NewResourceLogs()
+	rl.Resource().Attributes().PutStr(k8sLogType, "manifest")
+	sl := rl.ScopeLogs().AppendEmpty()
+	sl.Scope().Attributes().PutBool(otelEntityEventAsLog, true)
```
gantrior, line 36: I believe this should be set as resource attribute, see [the doc](https://swicloud.atlassian.net/wiki/spaces/SP/pages/4314595592/OTEL+Entity+Events+migration+guide)

gantrior (commented):
Few comments, but otherwise it looks good

src/processor/containerprocessor/log_constructor.go
janatrelakova: Correct me If I am wrong but based on the example:
```
{
  "resourceLogs": [
    {
      "scopeLogs": [
        {
          "scope": {
            "attributes": [
              {
                "key": "otel.entity.event_as_log",
                "value": {
                  "boolValue": true
                }
              }
            ]
          },
...
```
this attribute is a scope attribute.

gantrior: You are right, I am sorry for confusion. ok then.

src/processor/containerprocessor/processor.go
```diff
@@ -0,0 +1,96 @@
...
+		for j := range scopeLogs.Len() {
+			sl := scopeLogs.At(j)
+			logRecords := sl.LogRecords()
+
+			for k := range logRecords.Len() {
+				lr := logRecords.At(k)
+				attrs := lr.Attributes()
+
+				if !isPodLog(attrs) {
+					break
```
gantrior, line 65: you are existing for loop when one record isn't recognized as a Pod, is it intended? Shouldn't there be `continue` keyword instead?

src/processor/containerprocessor/processor.go
```diff
@@ -0,0 +1,96 @@
...
+			lr := containers.At(i)
+			lr.CopyTo(lrs.AppendEmpty())
+		}
+	}
+
+	return ld, nil
+}
+
+func (cp *containerprocessor) extractPodManifests(resourceLogs plog.ResourceLogsSlice) ([]Manifest, error) {
+	manifests := make([]Manifest, 0)
```
gantrior, line 50: Is it possible to streamline manifest processing by converting each manifest as you parse it and appending it to the logs immediatelly? It would avoid storing all manifests in a temporary slice. Something like:
```
for i := 0; i < resourceLogs.Len(); i++ {
    ...
    for j := 0; j < scopeLogs.Len(); j++ {
        ....
        for k := 0; k < logRecords.Len(); k++ {
            ....
            var m Manifest
            if err := json.Unmarshal([]byte(lr.Body().AsString()), &m); err != nil {
                cp.logger.Error("Error while unmarshalling manifest", zap.Error(err))
                return plog.NewLogs(), err
            }

            containers := transformManifestToContainerLogs(m)
            for ci := 0; ci < containers.Len(); ci++ {
                containerLog := containers.At(ci)
                containerLog.CopyTo(lrs.AppendEmpty())
            }
        }
    }
}
```

janatrelakova: I was looking at it and it felt like I was overloading responsibility of this function. I have decided to go with more go native approach and use unbuffered channel.

If you have any problem I can easily switch to the old way.

janatrelakova:
@gantrior I would like to double check `sidecar` question from the description as it may break some existing things.

Currently, `sw.k8s.container.init` is true for all `sidecar` containers. Is that alright with you?

Additionally, you can check an example structure of logs in the description.

gantrior (approved):

gantrior (commented):
I just looked at how `sw.k8s.container.sidecar` is evaluated and it is not in line with https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/876. For that you would need to follow this logic: set `sw.k8s.container.sidecar` to `True` if the container is init container and in the same time there is restartPolicy set to `Always`, `False` otherwise. CC: @pstranak-sw

janatrelakova:
> I just looked at how sw.k8s.container.sidecar is evaluated and it is not in line with https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/876. For that you would need to follow this logic: set sw.k8s.container.sidecar to True if the container is init container and in the same time there is restartPolicy set to Always, False otherwise.

@gantrior hmmm, when setting the sidecar property, I am going through spec -> initContainers. So I should have the guarantee that all containers in this collection are `init`. Additionally comparing restart policy to Always. 
So in case a container is in init collection and has restart policy set to Always, it is marked as a Sidecar.
```
for _, ic := range m.Spec.InitContainers {
		containers[ic.Name] = Container{
			Name:               ic.Name,
			IsInitContainer:    true,
			IsSidecarContainer: ic.RestartPolicy == "Always",
		}
	}
```

Or am I missing something?

gantrior:
> > I just looked at how sw.k8s.container.sidecar is evaluated and it is not in line with #876. For that you would need to follow this logic: set sw.k8s.container.sidecar to True if the container is init container and in the same time there is restartPolicy set to Always, False otherwise.
> 
> @gantrior hmmm, when setting the sidecar property, I am going through spec -> initContainers. So I should have the guarantee that all containers in this collection are `init`. Additionally comparing restart policy to Always. So in case a container is in init collection and has restart policy set to Always, it is marked as a Sidecar.
> 
> ```
> for _, ic := range m.Spec.InitContainers {
> 		containers[ic.Name] = Container{
> 			Name:               ic.Name,
> 			IsInitContainer:    true,
> 			IsSidecarContainer: ic.RestartPolicy == "Always",
> 		}
> 	}
> ```
> 
> Or am I missing something?

My bad, I didn't notice this part of the code. So it is ok then.

gantrior (dismissed):

etichy (approved):

etichy (approved):

PR #891, opened 2025-02-07
Remove obsolete swi-k8s-opentelemetry-collector

etichy:


index.yaml
pstranak-sw, line 1: Do we need to update the `generated` field, too?

pstranak-sw (approved):

PR #898, opened 2025-02-12
detect mainfest diffs

mpecenka:
forking k8sobjectsreceiver and modifying it to add tributes sw.metadata.changed, sw.spec.changed, sw.status.changed to logs/events with manifests
also added persistent storage for the receiver, to be able to keep data over restart


src/receiver/k8sobjectsreceiver/receiver.go
```diff
@@ -0,0 +1,491 @@
...
+	key     string
+	Objects map[string]objecthashes `json:"objects"`
+	mu      sync.Mutex
+}
+
+type objecthashes struct {
+	Metadata string `json:"metadata"`
+	Spec     string `json:"spec"`
+	Status   string `json:"status"`
+}
```
mpecenka, line 58: this is new code - added types for storing hashes to memory and storage

src/receiver/k8sobjectsreceiver/receiver.go
```diff
@@ -0,0 +1,491 @@
...
+		return "", err
+	}
+
+	return fmt.Sprintf("%x", sha256.Sum256(bytes)), nil
+}
+
+// key for the storage
+func getKey(udata *unstructured.Unstructured) string {
+	return fmt.Sprintf("%s:%s#%s", udata.GetKind(), udata.GetNamespace(), udata.GetName())
+}
```
mpecenka, line 402: there are new methods for calculating hashes from manifests and saving it to storage

PR #905, opened 2025-02-13
Fix timestamp of entity state events

janatrelakova:
## Summary
When initial poll is being done it goes through all manifests. The timestamp was previously filled with the last change in manifest (parsed from manifest's `conditions` part) and it could have been very old change as not every manifest changes frequently.  There are some restrictions in the ingestion pipeline ignoring log records outside of accepted time interval so it was ignoring containers created from manifests that did not update lately.

PR introducing new processor including this bug: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/878

## Solution
Timestamp of new Entity state event log is filled by timestamp of processed `logRecord` which caries `manifest` as in its body.

### DIFF
--- a/src/processor/containerprocessor/processor.go
+++ b/src/processor/containerprocessor/processor.go
@@ -136,7 +136,7 @@ func getTimestamp(lr plog.LogRecord) pcommon.Timestamp {
 		return lr.ObservedTimestamp()
 	}
 
-	if lr.Timestamp().AsTime().IsZero() {
+	if !lr.Timestamp().AsTime().IsZero() {
 		return lr.Timestamp()
 	}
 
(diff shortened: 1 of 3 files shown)


src/processor/containerprocessor/processor.go
```diff
@@ -113,13 +118,31 @@ func (cp *containerprocessor) generateManifests(mCh chan<- Manifest, errCh chan<
...
+func getTimestamp(lr plog.LogRecord) pcommon.Timestamp {
+	if !lr.ObservedTimestamp().AsTime().IsZero() {
+		return lr.ObservedTimestamp()
+	}
+
+	if !lr.Timestamp().AsTime().IsZero() {
+		return lr.Timestamp()
+	}
+
+	return pcommon.NewTimestampFromTime(time.Now())
```
janatrelakova, line 143: This is defensive approach, manifest logs are sent with `observedTimestamp` however it could be set in the `timestamp` property https://opentelemetry.io/docs/specs/otel/logs/data-model/#field-observedtimestamp.

gantrior (approved):

PR #916, opened 2025-02-21
Namespaces filter

mpecenka:
possibility to specify global filter for namespaces

### DIFF
--- a/deploy/helm/node-collector-config.yaml
+++ b/deploy/helm/node-collector-config.yaml
@@ -95,6 +95,18 @@ processors:
 {{ toYaml .Values.otel.metrics.filter | indent 6 }}
 {{- end }}
 
+{{- if eq (include "isNamespacesFilterEnabled" .) "true" }}
+  filter/namespaces:
+    metrics:
+      metric:
+{{- include "namespacesFilter" . | nindent 8 }}    
+
+  filter/namespaces_logs:
+    logs:
+      log_record:
+{{- include "namespacesFilter" . | nindent 8 }}      
+{{- end }}
+
 {{- include "common-config.filter-remove-temporary-metrics" . | nindent 2 }}
 
   {{- include "common-config.resource-metrics" . | nindent 2 }}
@@ -745,6 +757,9 @@ service:
         - groupbyattrs/common-all
         - resource/container
         - k8sattributes
+{{- if eq (include "isNamespacesFilterEnabled" .) "true" }}
+        - filter/namespaces_logs
+{{- end }}           
 {{- if eq (include "isDeprecatedFilterSyntax" .Values.otel.logs.filter) "false" }}
 {{- if (include "logsFilter" .) }}
         - filter/logs
@@ -838,6 +853,9 @@ service:
         - memory_limiter
         - filter/histograms
         - k8sattributes
+{{- if eq (include "isNamespacesFilterEnabled" .) "true" }}
+        - filter/namespaces
+{{- end }}             
 {{- if .Values.otel.metrics.filter }}
         - filter/metrics
 {{- end }}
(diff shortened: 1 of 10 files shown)


deploy/helm/templates/asserts.yaml
```diff
@@ -16,4 +16,8 @@
 
 {{- if eq ((semver "3.9.0").Compare (semver .Capabilities.HelmVersion.Version)) 1 -}}
 {{ fail "This version of Helm is not supported. Please use 3.9.0 or newer." }}
-{{- end -}}
\ No newline at end of file
+{{- end -}}
+
+{{- if and (or (not (empty .Values.cluster.filter.exclude_namespaces)) (not (empty .Values.cluster.filter.exclude_namespaces_regex))) (or (not (empty .Values.cluster.filter.include_namespaces)) (not (empty .Values.cluster.filter.include_namespaces_regex))) -}}
```
gantrior, line 21: Is this the same as `isNamespacesFilterEnabled` macro? can you use it here?

deploy/helm/values.yaml
```diff
@@ -649,6 +649,12 @@ cluster:
   # Useful for example if multiple clusters have to have the same name. Or when the name contains problematic characters.
   uid: ""
 
+  filter:
+    exclude_namespaces: []
+    exclude_namespaces_regex: []
+    include_namespaces: []
+    include_namespaces_regex: []
+
```
gantrior, line 657: `values.yaml` serves also as documentation of the options, so please describe the usage in the comment (especially the condition that only one of these can be set)

deploy/helm/templates/events-collector-deployment.yaml
```diff
@@ -34,7 +34,8 @@ spec:
     spec:
       terminationGracePeriodSeconds: {{ .Values.otel.events.terminationGracePeriodSeconds }}
       serviceAccountName: {{ include "common.fullname" . }}
-      securityContext: {}
+      securityContext: 
+        fsGroup: 10001
```
gantrior, line 38: How does this related to namespace filter?

deploy/helm/node-collector-config.yaml
```diff
@@ -95,6 +95,18 @@ processors:
 {{ toYaml .Values.otel.metrics.filter | indent 6 }}
 {{- end }}
 
+{{- if eq (include "isNamespacesFilterEnabled" .) "true" }}
+  filter/namespaces:
+    metrics:
+      metric:
+{{- include "namespacesFilter" . | nindent 8 }}
```
gantrior, line 102: Have you tested this? You are using `metric` context of filterprocessor which is meant to filter whole metrics out, but we meant to filter just individual datapoints so I guess you should be using `datapoint` context (see https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor#configuration)

gantrior (changes requested):
I have some comments that should be addressed first

deploy/helm/templates/asserts.yaml
mpecenka: it is not the same

deploy/helm/templates/events-collector-deployment.yaml
mpecenka: it is not, that is the fix for the storage issue

deploy/helm/node-collector-config.yaml
mpecenka: hmm I do not know, it is written the same way as our current filters, you just do not have to specify multiple filters for metrics, logs, events, but the usage in pipeline is the same

mpecenka: current values.yaml comments suggest using metric for filter metrics by namespace:

    # This filter is applied after metric processing, it is the place where metrics could be filtered out
    # https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/filterprocessor for configuration reference
    # With filter conditions we build a global blacklist. So if we exclude data, let's say by specific namespace,
    # we need to allow other data to be collected. The following example excludes all data from namespaces other than 'foo',
    # but we need to add a condition that allows data without namespace attribute to be collected.
    # filter:
    #   metric:
    #     - not(IsMatch(resource.attributes["k8s.namespace.name"], "foo") or resource.attributes["k8s.namespace.name"] == nil)
    filter: {}

gantrior (approved):

PR #918, opened 2025-02-24
Adding support for Prometheus CRDs

gantrior:
* Depends on https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/919 (contains also its diff)
* Adding `crd` subchart which contains Prometheus Operator CRDs
    * Adding CRDs to our heml chart recommended by [Otel Operator doc](https://github.com/open-telemetry/opentelemetry-operator/tree/main/cmd/otel-allocator):
    > The easiest way to do this is to grab a copy of the individual [PodMonitor](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/charts/crds/crds/crd-podmonitors.yaml) YAML and [ServiceMonitor](https://github.com/prometheus-community/helm-charts/blob/main/charts/kube-prometheus-stack/charts/crds/crds/crd-servicemonitors.yaml) YAML custom resource definitions (CRDs) from the [Kube Prometheus Operator’s Helm chart](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/charts).
    * the reason is that we need it to be in `crds` Helm folder so that it is applied before anything else, but in `crds` folder in cannot be conditional. Therefore we add it as subchart where whole subchart is covered with the condition. 
* For dev purposes installing Prometheus Operator instead of raw prometheus. This brings some default ServiceMonitors with which the chart can be tested. 
* Updated Otel Operator to the version that matches current version of Otel Collector (`0.116.0`)
* Deploying `OpenTelemetryCollector` CR in `statefulset` mode with TargetAllocator enabled and configured to scrape only Prometheus endpoints
    * Pipeline is 90% same as node_collector's discovery pipeline, with few minor modifications
    * `common.fullname` was modified to be able to define custom max_length as StatefulSet in kubernetes must be at most 52 characts to be able to deploy Pods (see https://github.com/kubernetes/kubernetes/issues/64023)
    * Adding default `autoscaler` option (configurable through `values.yaml`)
    * Adding default options for `serviceMonitorSelector`, `podMonitorSelector`, `probeSelector` and `scrapeConfigSelector` (configurabled through `values.yaml`)
    * Added `wait-for-certmanager` that runs when `certmanager.enabled` is set to `true`
    * Added `wait-for-operator` that runs when `operator.enabled` is set to `true`
    * `imagePullSecrets` is set to ServiceAccount as there is no configuration for it in `OpenTelemetryCollector` CR
* Bumping as `4.4.0-alpha.15`

### DIFF
--- a/deploy/helm/templates/operator/discovery-collector.yaml
+++ b/deploy/helm/templates/operator/discovery-collector.yaml
@@ -1,0 +1,487 @@
+{{- if .Values.otel.metrics.autodiscovery.discovery_collector.enabled }}
+apiVersion: opentelemetry.io/v1beta1
+kind: OpenTelemetryCollector
+metadata:
+  # name is limitted to 41, as `-collector` is autoadded by the operator and statefulset can have maximum 52 characters
+  name: {{ include "common.fullname" (tuple . "-discovery" 41) }} 
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "common.labels" . | indent 4 }}
+  annotations:
+    "helm.sh/hook": post-install,post-upgrade
+    "helm.sh/hook-weight": "2"
+spec:
+  mode: statefulset
+  autoscaler:
+    {{- if .Values.otel.metrics.autodiscovery.discovery_collector.autoscaler }}
+{{ toYaml .Values.otel.metrics.autodiscovery.discovery_collector.autoscaler | nindent 4 }}
+    {{- else }}
+    minReplicas: 1
+    maxReplicas: 5
+    targetCPUUtilization: 80
+    targetMemoryUtilization: 80
+    {{- end }}
+  targetAllocator:
+    enabled: true
+    serviceAccount: {{ include "common.fullname" . }}
+    prometheusCR:
+      enabled: true
+      scrapeInterval: {{ quote .Values.otel.metrics.prometheus.scrape_interval }}
+      serviceMonitorSelector: 
+        {{- if .Values.otel.metrics.autodiscovery.discovery_collector.targetAllocator.serviceMonitorSelector }}
+{{ toYaml .Values.otel.metrics.autodiscovery.discovery_collector.targetAllocator.serviceMonitorSelector | nindent 8 }}
+        {{- else }}
+        matchExpressions: 
+          - key: sw.ignore
+            operator: NotIn
+            values:
+            - "true"
+        {{- end }}
+      podMonitorSelector: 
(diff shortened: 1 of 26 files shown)


deploy/helm/templates/operator/discovery-collector.yaml
```diff
@@ -0,0 +1,487 @@
...
+    serviceAccount: {{ include "common.fullname" . }}
+    prometheusCR:
+      enabled: true
+      scrapeInterval: {{ quote .Values.otel.metrics.prometheus.scrape_interval }}
+      serviceMonitorSelector: 
+        {{- if .Values.otel.metrics.autodiscovery.discovery_collector.targetAllocator.serviceMonitorSelector }}
+{{ toYaml .Values.otel.metrics.autodiscovery.discovery_collector.targetAllocator.serviceMonitorSelector | nindent 8 }}
+        {{- else }}
+        matchExpressions: 
+          - key: sw.ignore
```
etichy, line 35: Why sw.ignore NotIn true vs sw.ignore In false?

gantrior: I wanted to cover all ServiceMonitors that don't even have any `sw.ignore` label defined, so basically just exclude those that have specificly `sw.ignore=true` label.

deploy/helm/templates/operator/wait-for-credmanager.yaml
```diff
@@ -0,0 +1,30 @@
...
+            - |
+              while [ "$(kubectl get pods -n {{ .Release.Namespace }} -l app.kubernetes.io/name=certmanager \
+                -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}')" != "True" ]; do
+                kubectl get pods -n {{ .Release.Namespace }} -l app.kubernetes.io/name=certmanager
+                kubectl get pods -n {{ .Release.Namespace }} -l app.kubernetes.io/name=certmanager -o 'jsonpath={..status.conditions[?(@.type=="Ready")].status}'
+                echo "Waiting for certmanager to be ready..."
+                sleep 5
+              done
+      restartPolicy: Never
+  backoffLimit: 4
```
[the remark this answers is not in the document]
gantrior: ok it makes sense to define the limit, I defined `activeDeadlineSeconds: 300` which should cover all retries all together

deploy/helm/templates/operator/discovery-collector.yaml
etichy: make sense

etichy (approved):

PR #930, opened 2025-02-26
vulnerability fix

mpecenka:
fixing golang.org/x/net vulnerability


src/cmd/go.mod
```diff
@@ -306,7 +306,7 @@ require (
 	golang.org/x/crypto v0.31.0 // indirect
 	golang.org/x/exp v0.0.0-20240506185415-9bf2ced13842 // indirect
 	golang.org/x/mod v0.19.0 // indirect
-	golang.org/x/net v0.32.0 // indirect
+	golang.org/x/net v0.33.0 // indirect
```
gantrior, line 309: Directly updating version indirect depencies is not ideal. If I simply run `go mod tidy` don't I cleanup your changes? Why don't you use `replace` section, which is the way how we did security fixes in the past. E.g.:
```
replace golang.org/x/net v0.32.0 => golang.org/x/net v0.33.0
```

gantrior: here is PR where you were fixing it using `replace` section: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/290/files

mpecenka: I did run `go mod tidy` after changing go.mod.

gantrior: anyway using `replace` section is more robust so I would go with that.

pstranak-sw: @mpecenka did it correctly. That being said, using `replace` only in the `src/cmd/go.mod` would work, too.

pstranak-sw (approved):

PR #879, opened 2025-01-30
NH-99740: Removing docker image handling

gantrior:
Docker image is taken from https://github.com/solarwinds/solarwinds-otel-collector where we moved all the code

* For development purposes (in CI builds and locally) we will still do builds of the docker image from source (added https://github.com/solarwinds/solarwinds-otel-collector as submodule), once we stabilise development of k8s changes this may be removed in the future. 
    * submodule will be updated by dependabot on daily basis
* It by default we use image tag with `-k8s` which has limited subset of capabilities designed for our purpose.

TODOs:
* [x] get rid of swik8sattributes processor (or move it to https://github.com/solarwinds/solarwinds-otel-collector) - https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/872
* [x] get new release of solarwinds-otel-collector which includes:
    * https://github.com/solarwinds/solarwinds-otel-collector/pull/33
    * https://github.com/solarwinds/solarwinds-otel-collector/pull/32
    * https://github.com/solarwinds/solarwinds-otel-collector/pull/36
* [x] wait for  https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/878 to be merged and then move it to https://github.com/solarwinds/solarwinds-otel-collector
* [x] get rid of for of k8sobjectsreceiver
    * https://github.com/solarwinds/solarwinds-otel-collector/pull/74
    * https://github.com/solarwinds/solarwinds-otel-collector/pull/73
* [x] get rid of wrapper in solarwinds-otel-collector: https://github.com/solarwinds/solarwinds-otel-collector/pull/61
* [x] migrate containerprocessor to solarwinds-otel-collector: https://github.com/solarwinds/solarwinds-otel-collector/pull/60
* [x] wait for all features to be release in the docker image - https://github.com/solarwinds/solarwinds-otel-collector/pull/75

PR #947, opened 2025-03-05
Update Helm unit tests to be compatible with the new Helm unittest plugin

pstranak-sw:
The `unittest` plugin in version 0.8.0 changed how test skipping works.


etichy (dismissed):

deploy/helm/tests/node-collector-config-map-windows_test.yaml
```diff
@@ -1,6 +1,7 @@
 # yaml-language-server: $schema=https://raw.githubusercontent.com/helm-unittest/helm-unittest/main/schema/helm-testsuite.json
 suite: Test for node-collector-config-map-windows
-skip: true # Unresolved problem with backslash 
+skip:
+  reason: Unresolved problem with backslash
```
gantrior, line 4: Is this reason still valid? If I removed the skip (I did it in [my PR](https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/879)), the tests are working just fine

pstranak-sw: I haven't checked. My goal was only to fix the currently failing build pipeline.

etichy: no clue and really it works on my local windows box now. Maybe we removed some filters or configuration which causes troubles? Anyway you can remove that skip

etichy (approved):

PR #976, opened 2025-03-31
NH-104532 fix log warnings

jakub-racek-swi:


deploy/helm/templates/operator/discovery-collector.yaml
```diff
@@ -245,13 +245,13 @@ spec:
         metric_statements:
           - context: scope
```
pstranak-sw, line 246: We don't need to set the context anymore, right?

jakub-racek-swi: that is correct, I'll update it

jakub-racek-swi: nevermind actually, it needs to set the context to find the path, and crashes otherwise

deploy/helm/templates/operator/discovery-collector.yaml
pstranak-sw, line 1: This file also contains the `service.telemetry:` section.

deploy/helm/metrics-collector-config.yaml
```diff
@@ -1170,5 +1170,10 @@ service:
 {{- end }}
 {{- if .Values.otel.metrics.telemetry.metrics.enabled }}
     metrics:
-      address: {{ .Values.otel.metrics.telemetry.metrics.address }}
+      readers:
+        - pull:
+            exporter:
+              prometheus:
+                host: {{ (split ":" .Values.otel.metrics.telemetry.metrics.address)._0 | quote }}
```
pstranak-sw, line 1177: I'm not fan of the way we pass these telemetry settings to the OTEL Collector. I'd rather pass them as-is. And use this parsing logic only when the customer used the deprecated `address` setting. We even say that the documentation for the available telemetry settings is the official one from the OTEL Collector.

deploy/helm/templates/operator/discovery-collector.yaml
pstranak-sw: The OTEL Operator does not support this new setting, yet. We have to leave this as-is for now.

deploy/helm/node-collector-config.yaml
```diff
@@ -216,8 +216,8 @@ processors:
           - set( attributes["host.name"], attributes["k8s.pod.name"])
```
pstranak-sw, line 216: Why not update also these and remove the context from line 213?

deploy/helm/node-collector-config.yaml
```diff
@@ -289,23 +289,21 @@ processors:
...
-      - context: scope
-        statements:
-          - set(name, "")
-          - set(version, "")
+      - statements:
+        - set(scope.name, "")
+        - set(scope.version, "")
   transform/istio-metrics:
     metric_statements:
       - context: metric
```
pstranak-sw, line 301: Is this context still needed?

deploy/helm/templates/_common-config.tpl
```diff
@@ -34,7 +34,7 @@ transform/unify_node_attribute:
     - context: datapoint
```
pstranak-sw, line 34: Is this context still needed?

deploy/helm/templates/operator/discovery-collector.yaml
```diff
@@ -310,10 +308,10 @@ spec:
         metric_statements:
           - context: metric
```
pstranak-sw, line 309: Is this context still needed?

pstranak-sw (changes requested):
When I deploy the k8s collector with these changes, I still see several log messages about settings that need to be updated.

jakub-racek-swi:
@pstranak-sw not sure how I missed so many changes last time, all the redundant context statements should be gone from metric and event configurations now and no more warnings showed up when testing on docker-desktop

pstranak-sw (approved):

PR #981, opened 2025-04-02
adding Beyla as optional part of the chart

gantrior:
* it is opt-in using `beyla.enable`
* it does not reference subchart, I added beyla's resources directly to our chart, so that we don't need to configure it through default values.yaml

### DIFF
--- a/deploy/helm/templates/beyla/configmap.yaml
+++ b/deploy/helm/templates/beyla/configmap.yaml
@@ -1,0 +1,42 @@
+{{- if .Values.beyla.enabled }}
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: {{ include "common.fullname" (tuple . "-beyla-config") }}
+  namespace: {{ .Release.Namespace }}
+  labels:
+{{ include "common.labels" . | indent 4 }}
+    app.kubernetes.io/component: config
+  annotations:
+{{ include "common.annotations" . | indent 4 }}
+data:
+  beyla-config.yml: |
+    otel_metrics_export:
+      endpoint: http://{{ include "common.fullname" (tuple . "-metrics-collector") }}:{{ .Values.otel.metrics.otlp_endpoint.port }}
+      protocol: grpc
+      interval: {{ quote .Values.otel.metrics.prometheus.scrape_interval }}
+      features: ["application", "application_process", "application_service_graph", "network"]
+    ebpf:
+      enable_context_propagation: true
+    process:
+      enabled: true
+    network:
+      enable: true
+    discovery:
+      services:
+        - k8s_namespace: .*
+      exclude_services:
+        - exe_path: ".*otelcol.*|.*beyla.*"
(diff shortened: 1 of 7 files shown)


deploy/helm/templates/beyla/daemon-set.yaml
pstranak-sw, line 1: Is this compatible with OpenShift?

deploy/helm/templates/beyla/daemon-set.yaml
```diff
@@ -0,0 +1,110 @@
...
+  template:
+    metadata:
+      annotations:
+        checksum/config: {{ include (print $.Template.BasePath "/beyla/configmap.yaml") . | sha256sum }}
+{{ include "common.annotations" . | indent 8 }}
+      labels:
+        app: {{ include "common.fullname" (tuple . "-beyla") }}
+{{ include "common.labels" . | nindent 8 }}
+{{ include "common.pod-labels" . | indent 8 }}
+        app.kubernetes.io/component: workload
```
pstranak-sw, line 24: What's the purpose of this label?

deploy/helm/templates/beyla/configmap.yaml
```diff
@@ -0,0 +1,41 @@
...
+    attributes:
+      kubernetes:
+        enable: true
+    prometheus_export:
+      port: 9090
+      path: /metrics
+    internal_metrics:
+      prometheus:
+        port: 6060
+        path: /metrics
```
pstranak-sw, line 40: Are these temporary for alphas. Or do we want to keep them also once released?

deploy/helm/templates/beyla/daemon-set.yaml
gantrior: I haven't tested it, it does not run in privileged mode so it should be working.

deploy/helm/templates/beyla/daemon-set.yaml
gantrior: There is no purpose, it was in original resource in beyla's chart, so I left it there

deploy/helm/templates/beyla/configmap.yaml
gantrior: I would remove them later, I would like to see if there is anything useful there

pstranak-sw (approved):

deploy/helm/values.schema.json
```diff
@@ -1521,6 +1521,34 @@
...
+                    "$ref": "#/definitions/imageDefinition"
+                },
+                "nodeSelector": {
+                    "$ref": "#/definitions/k8sNodeSelector"
+                },
+                "resources": {
+                    "$ref": "#/definitions/k8sResourceRequirements"
+                },
+                "telemetry": {
+                    "$ref": "#/definitions/ebpfTelemetry"
```
pstranak-sw, line 1545: I'd prefer not reusing ebpf stuff here.

PR #1024, opened 2025-05-16
Add github action to update images in values.yaml

murad-sw:
This GitHub Action script automatically updates Docker image versions in Helm chart files (`values.yaml` and `Chart.yaml`).

- Scans the values.yaml file for Docker image configurations with repository and tag fields
- Fetches the latest semantic versions from Docker registries (Docker Hub, GHCR)
- Updates image tags to their latest versions if newer ones are available
- Bumps the Helm chart version and updates the `appVersion`
- Creates/updates a Git branch and PR with the changes

STILL manual *close and open PR* needed to run builds


utils/update_images.py
```diff
@@ -0,0 +1,329 @@
...
+def get_latest_tag(registry, repository, version_pattern=None, github_token=None):
+    """Get the latest tag for an image"""
+    if registry == "docker.io":
+        tags = get_docker_hub_tags(repository, version_pattern)
+        return get_latest_version(tags)
+    elif registry == "ghcr.io":
+        from urllib.parse import urlparse
+        parsed_url = urlparse(repository)
+        if parsed_url.hostname == "ghcr.io":
+            repo_name = parsed_url.path.lstrip("/")
```
gantrior, line 98: What if the repo is `ghcr.io/owner/img`? It seems that end up being `https://ghcr.io/v2/ghcr.io/owner/img/tags/list` with double repo in it.

utils/update_images.py
```diff
@@ -0,0 +1,329 @@
...
+        api_repository = f"library/{repository}"
+    else:
+        api_repository = repository
+    
+    next_url = f"https://hub.docker.com/v2/repositories/{api_repository}/tags?page_size=100"
+    logger.info(f"Fetching tags from: {next_url}")
+    
+    while next_url:
+        try:
+            response = requests.get(next_url)
```
gantrior, line 28: All `requests.get` calls should have timeout, to not hang it indefinitely

utils/update_images.py
```diff
@@ -0,0 +1,329 @@
...
+            chart_data["version"] = new_chart_version
+            
+            if app_version:
+                chart_data["appVersion"] = app_version
+            
+            with open(chart_file, 'w') as f:
+                yaml.dump(chart_data, f)
+            
+            logger.info(f"Updated Chart.yaml: version {chart_version} -> {new_chart_version}")
+            return True
```
gantrior, line 253: It does not make sense to return anything, it is not used anyway and function returns `True` only on the happy path, but never returns `False`. It is better to always return boolean and react on that on the caller though.

utils/update_images.py
```diff
@@ -0,0 +1,329 @@
...
+    
+    for change in changes:
+        path = change["path"]
+        new_tag = change["new_tag"]
+        
+        current = data
+        for part in path[:-1]:
+            if part not in current:
+                logger.warning(f"Path {path} not found in YAML")
+                break
```
gantrior, line 204: This breaks the loop but continues with the next change, leaving current in an undefined state. You should have `continue` here.

utils/update_images.py
murad-sw: you're right, I'm not sure how I need to handle situations when this function returns False, so I better remove this

I probably had some condition there, but later removed it

utils/update_images.py
murad-sw: I clean repository before calling `get_ghcr_tags()`:

```
    elif registry == "ghcr.io":
        parsed_url = urlparse(repository)
        if parsed_url.hostname == "ghcr.io":
            repo_name = parsed_url.path.lstrip("/")
        else:
            repo_name = repository
```

gantrior (dismissed):

gantrior (approved):

PR #1043, opened 2025-05-28
NH-108817: Creating relationship state events from Istio metrics

gantrior:
I added several pipelines to Node Collector, Metrics Discovery Collector and Discovery Collector (see `_common-discovery-config.tpl` for definitions):

* metrics/discovery-custom
* metrics/discovery-scrape (this replaces previous metrics/discovery)
* metrics/discovery-istio
* metrics/relationship-state-events-workload-workload-preparation
* metrics/relationship-state-events-workload-service-preparation
* logs/stateevents

This uses [solarwindsentityconnector](https://github.com/solarwinds/solarwinds-otel-collector-contrib/tree/main/connector/solarwindsentityconnector)

as the results there should be stream of relationship state events sent based on `istio_request_bytes_sum` metric, which should produce following relationships:
* Workload -> Workload
* Workload -> Service

### DIFF
--- a/deploy/helm/metrics-discovery-config.yaml
+++ b/deploy/helm/metrics-discovery-config.yaml
@@ -42,15 +42,6 @@ processors:
       - statements:
         - set(scope.name, "")
         - set(scope.version, "")
-
-  metricstransform/rename/discovery:
-    transforms:
-      # add `k8s.` prefix to all metrics
-      - include: ^(.*)$$
-        match_type: regexp
-        action: update
-        new_name: {{ .Values.otel.metrics.autodiscovery.prefix }}$${1}
-
  
   groupbyattrs/common-all:
     keys:
@@ -98,7 +89,7 @@ processors:
 
   batch/metrics:
 {{ toYaml .Values.aws_fargate.metrics.autodiscovery.batch | indent 4 }}
-  
+
 {{- include "common-config.filter-remove-temporary-metrics" . | nindent 2 }}
 
   filter/histograms:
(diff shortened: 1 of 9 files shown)


deploy/helm/node-collector-config.yaml
```diff
@@ -1,4 +1,9 @@
 exporters:
+  debug:
```
pstranak-sw, line 2: Should this stay here?

deploy/helm/templates/_common-discovery-config.tpl
```diff
@@ -0,0 +1,505 @@
...
+  workload_mappings:
+    - name_attr: source_workload
+      namespace_attr: source_workload_namespace
+      workload_type_attr: source_workload_type
+      expected_types:
+        - deployments
+        - daemonsets
+        - statefulsets
+        - jobs
+        - cronjobs
```
pstranak-sw, line 153: Do we actually get a CronJob as a workload in Istio?

deploy/helm/metrics-discovery-config.yaml
```diff
@@ -98,74 +89,21 @@ processors:
...
-  metricstransform/copy-required-metrics:
-    transforms:
-    {{- $root := . }}
-    {{- range $index, $metric := $arrayOfRequiredMetrics }}
-      - include: {{ $root.Values.otel.metrics.autodiscovery.prefix }}{{ $metric }}
-        action: insert
-        new_name: k8s.{{ $metric }}
-    {{- end }}
-{{- end }}
+{{ include "common-discovery-config.processors" . | indent 2 }}
```
pstranak-sw, line 100: This template includes processors for Istio metrics. Do we get them also in Fargate environment?

gantrior: Istio is probably not supported on Fargate, but if it is mixed cluster the istio can run on standard nodes so it makes sense to process that.

deploy/helm/node-collector-config.yaml
gantrior: Removing, forgotten

deploy/helm/templates/_common-discovery-config.tpl
```diff
@@ -0,0 +1,505 @@
...
+        - pods
```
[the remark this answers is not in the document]
gantrior: no we don't I am removing it.

pstranak-sw (approved):

PR #1050, opened 2025-05-29
Check for cert-manager when enabling deployment of OTEL operator

pstranak-sw:
If `operator.enabled` is set to `true`, check if `certmanager.enabled` is also `true` or if there is an already running `cert-manager` in the cluster.

### DIFF
--- a/deploy/helm/templates/asserts.yaml
+++ b/deploy/helm/templates/asserts.yaml
@@ -21,3 +21,17 @@
 {{- if and (or (not (empty .Values.cluster.filter.exclude_namespaces)) (not (empty .Values.cluster.filter.exclude_namespaces_regex))) (or (not (empty .Values.cluster.filter.include_namespaces)) (not (empty .Values.cluster.filter.include_namespaces_regex))) -}}
 {{ fail "Only one namespace filter can be used at time, either include or exclude." }}
 {{- end -}}
+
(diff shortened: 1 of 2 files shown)


deploy/helm/templates/asserts.yaml
```diff
@@ -21,3 +21,17 @@
 {{- if and (or (not (empty .Values.cluster.filter.exclude_namespaces)) (not (empty .Values.cluster.filter.exclude_namespaces_regex))) (or (not (empty .Values.cluster.filter.include_namespaces)) (not (empty .Values.cluster.filter.include_namespaces_regex))) -}}
 {{ fail "Only one namespace filter can be used at time, either include or exclude." }}
 {{- end -}}
+
+{{- if and .Values.operator.enabled .Values.operator.admissionWebhooks.certManager.enabled (not .Values.certmanager.enabled) -}}
+{{- $certManagerFound:=false -}}
+{{- range $index, $conf := (lookup "admissionregistration.k8s.io/v1" "MutatingWebhookConfiguration" "" "").items -}}
+{{- range $index, $webhook := $conf.webhooks -}}
+{{- if eq $webhook.name "webhook.cert-manager.io" -}}
```
pstranak-sw, line 29: We cannot check for a specific workload or CR, because they are only identifiable by labels, not by names. And we also cannot check for a specific CRD, because they are left in the cluster even when the `cert-manager` is uninstalled. So we are checking for a specific webhook.

gantrior (approved):

PR #1063, opened 2025-06-23
Create relationships based on OTEL network metrics

pstranak-sw:
When Beyla is enabled (or standard OTEL Network metrics are sent to the gateway deployment), use the `swok8sworkloadtype` processor and `solarwindsentity` connector to detect network relationships between k8s resources.

Depends on https://github.com/solarwinds/solarwinds-otel-collector-contrib/pull/29 and release of a new image.


gantrior (dismissed):

deploy/helm/gateway-collector-config.yaml
```diff
@@ -97,32 +97,19 @@ processors:
...
 
   swok8sworkloadtype/beyla:
     workload_mappings:
       - name_attr: k8s.src.owner.name
         namespace_attr: k8s.src.namespace
-        workload_type_attr: k8s.src.owner.type
+        workload_type_attr: sw.k8s.src.workload.type
+        workload_name_attr: sw.k8s.src.workload.name
+        workload_namespace_attr: sw.k8s.src.workload.namespace
+        prefer_owner_for_pods: true
```
etichy, line 112: What about to have both, pod and also pod owner?

deploy/helm/gateway-collector-config.yaml
```diff
@@ -134,7 +121,10 @@ processors:
           - nodes
       - name_attr: k8s.dst.owner.name
         namespace_attr: k8s.dst.namespace
-        workload_type_attr: k8s.dst.owner.type
+        workload_type_attr: sw.k8s.dst.workload.type
+        workload_name_attr: sw.k8s.dst.workload.name
+        workload_namespace_attr: sw.k8s.dst.workload.namespace
+        prefer_owner_for_pods: true
```
etichy, line 127: Beyla attributes k8s.dst.owner.name and  k8s.dst.owner.type don't provide same information?

deploy/helm/gateway-collector-config.yaml
pstranak-sw: It could be configured that way. But we don't need that now, or do we?

deploy/helm/templates/beyla/configmap.yaml
```diff
@@ -50,6 +50,38 @@ data:
...
+            - rpc.system
+        beyla_*:
+          include:
+            - direction
+            - dst.address
+            - k8s.dst.namespace
+            - k8s.dst.owner.name
+            - src.address
+            - k8s.src.namespace
+            - k8s.src.owner.name
```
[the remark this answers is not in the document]
pstranak-sw: These come from https://grafana.com/docs/beyla/latest/network/#metric-attributes.
We are currently not interested in `k8s.src.type` because we only want the parent. And `k8s.src.owner.type` is currently limited to only a few types and Jobs and other less common ones are not supported by Beyla, yet.

We could ask for these attributes and then have some logic to decide whether they have values that we need or whether we need to ignore them and get the data ourselves. But it seemed to me that it would be simpler to just ignore them and always get the data ourselves.

deploy/helm/gateway-collector-config.yaml
pstranak-sw: Sometimes. See https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/1063/files/6a757758ddd8fcc3dde43374436e77e8050e22c0#r2167314257

deploy/helm/gateway-collector-config.yaml
```diff
@@ -156,6 +149,91 @@ processors:
...
+        expected_types:
+          - services
+          - pods
+
+  transform/beyla-fqdn-attribute:
+    metric_statements:
+      - statements:
+          # The logic of what shapes of addresses are considered FQDNs should match 'transform/istio-metric-datapoints'.
+          # Also, Beyla prefers filling 'client.address' and 'server.address' attributes with OTEL Service name (or name + k8s namespace) instead of an actual FQDN, when possible. This makes their values useless for us because they would create fake FQDNs.
```
pstranak-sw, line 175: This is unfortunate as it severely limits usefulness of Beyla for establishing HTTP/gRPC relationships between clients and servers. It was implemented that way intentionally. Though personally, I don't believe the reasoning there is correct.

deploy/helm/templates/beyla/configmap.yaml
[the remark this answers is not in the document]
etichy: ok, agree. Did we consider contribute to beyla so it supports all types?

PR #1070, opened 2025-07-08
Add devcontainer that connects to local docker-desktop

gantrior:


.devcontainer/local-features/copy-kube-config/devcontainer-feature.json
```diff
@@ -0,0 +1,11 @@
+{
+	"id": "copy-kube-config",
```
pstranak-sw, line 2: `version` is also mandatory

.devcontainer/local-features/copy-kube-config/copy-kube-config.sh
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash -i
+
+# Copies localhost's ~/.kube/config file into the container and swap out localhost
+# for kubernetes.docker.internal whenever a new shell starts to keep them in sync.
+if [ "$SYNC_LOCALHOST_KUBECONFIG" = "true" ] && [ -d "/usr/local/share/kube-localhost" ]; then
+    mkdir -p $HOME/.kube
+    sudo cp -r /usr/local/share/kube-localhost/* $HOME/.kube
```
pstranak-sw, line 7: Consider excluding `cache` subfolder. When there are many files in it on the host, it can block opening each new terminal for a few minutes.

PR #1098, opened 2025-08-08
Setting `max_staleness` property of cumulativetodelta to avoid memory leaks

gantrior:
* This is the result of analysis of `heap.out` of one of our customer. It turned out that cumulativetodelta is eating huge amount of memory, probably due to leak of cardinality and that it is keeping old cardinality in memory indefinitely. This should be avoided by setting `max_staleness` - see https://github.com/open-telemetry/opentelemetry-collector-contrib/tree/main/processor/cumulativetodeltaprocessor
> max_staleness: The total time a state entry will live past the time it was last seen. Set to 0 to retain state indefinitely. Default: 0

* setting max_staleness by default to triple of scrape_interval, making it configurable.
* bumping as bugfix release as 4.7.2

### DIFF
--- a/deploy/helm/templates/_helpers.tpl
+++ b/deploy/helm/templates/_helpers.tpl
@@ -380,7 +380,7 @@ Usage:
 
 {{/*
 Calculate max_staleness for cumulativetodelta processor
-If max_staleness is explicitly set, use it. Otherwise, default to twice the scrape_interval
+If max_staleness is explicitly set, use it. Otherwise, default to thrice the scrape_interval
 Input should be the Values object with prometheus configuration
 Output will be a duration string like "120s", "4m", etc.
 */}}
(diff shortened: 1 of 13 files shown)


juraj-paska (approved):

deploy/helm/templates/_helpers.tpl
```diff
@@ -378,6 +378,43 @@ Usage:
 {{- end -}}
 {{- end -}}
 
+{{/*
+Calculate max_staleness for cumulativetodelta processor
+If max_staleness is explicitly set, use it. Otherwise, default to twice the scrape_interval
```
juraj-paska, line 383:
```suggestion
If max_staleness is explicitly set, use it. Otherwise, default to thrice the scrape_interval
```

juraj-paska (approved):

PR #1144, opened 2025-10-02
NH-116356: Fix for missing external connections

gantrior:


deploy/helm/gateway-collector-config.yaml
```diff
@@ -205,8 +205,10 @@ processors:
...
-          # - set(datapoint.attributes["dest.sw.server.address.fqdn"], datapoint.attributes["server.address"]) where IsMatch(metric.name, "^http\\.client\\.") and datapoint.attributes["sw.k8s.dst.workload.type"] == nil and IsMatch(datapoint.attributes["server.address"], "^(https?://)?[a-zA-Z0-9][-a-zA-Z0-9]*\\.[a-zA-Z0-9][-a-zA-Z0-9\\.]*(:\\d+)?$") and not(IsMatch(datapoint.attributes["server.address"], ".*\\.cluster\\.local$")) and not(IsMatch(datapoint.attributes["server.address"], "^(https?://)?\\d+\\.\\d+\\.\\d+\\.\\d+(:\\d+)?$"))
+          - set(datapoint.attributes["dest.sw.server.address.fqdn"], datapoint.attributes["server.address"]) where IsMatch(metric.name, "^http\\.client\\.") and datapoint.attributes["sw.k8s.dst.workload.type"] == nil and IsMatch(datapoint.attributes["server.address"], "^(https?://)?[a-zA-Z0-9][-a-zA-Z0-9]*\\.[a-zA-Z0-9][-a-zA-Z0-9\\.]*(:\\d+)?$") and not(IsMatch(datapoint.attributes["server.address"], ".*\\.cluster\\.local$")) and not(IsMatch(datapoint.attributes["server.address"], "^(https?://)?\\d+\\.\\d+\\.\\d+\\.\\d+(:\\d+)?$"))
```
pstranak-sw, line 208: What will this do when the address is `some.address`, where `some` is an OTEL service and `address` is an OTEL namespace of the service?

pstranak-sw: Also, if you are checking for `.cluster.local`, you might also want to check for `.cluster.local.`.

gantrior: > Also, if you are checking for .cluster.local, you might also want to check for .cluster.local..
Fixed
> Additionally, for new workloads, sw.k8s.dst.workload.type might be temporary empty until k8s workload metadata cache gets updated. Or the swok8sworkloadtype processor may not be configured to look for that kind of a k8s object. In that case server.address may contain basically anything, including k8s workloads.
It is very unlikely that `server.address` will be filled with some new workload name before `swok8sworkloadtype` has that in cache, that would mean that beyla catched that new workload much earlier then `swok8sworkloadtype` (beyla needs to cache the workload and wait for next metric reporting interval), although it is teoretically possible. But I wouldn't block this PR with this.

PR #1165, opened 2025-10-22
Deploy the gateway-collector's HPA only when there is a Metrics Server in the cluster

pstranak-sw:


deploy/helm/templates/gateway/gateway-hpa.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.otel.gateway.enabled }}
+{{- if and .Values.otel.gateway.enabled (.Capabilities.APIVersions.Has "metrics.k8s.io/v1beta1") }}
```
gantrior: I am a bit afraid of overusing these kind of conditions. What if the version stabilizes on `metrics.k8s.io/v1`? For example Sumologic has it protected only by config values: https://github.com/SumoLogic/sumologic-kubernetes-collection/blob/89847b90aad8dbfbad0e50f34d2fc24f86e4064d/deploy/helm/sumologic/templates/logs/common/hpa.yaml#L1

Can we just introduce `otel.gateway.autoscaler.enabled` option to be able to disable autoscaler?

pstranak-sw: Sure, that's a rather common pattern. But in that case it would have to be disabled by default.

gantrior (approved):

PR #1168, opened 2025-10-23
Allow running the SWO Agent on ARM Nodes

pstranak-sw:


deploy/helm/templates/swo-agent-statefulset.yaml
```diff
@@ -37,16 +37,27 @@ spec:
...
         {{ toYaml .Values.swoagent.nodeSelector | nindent 8 }}
-      {{- else }}
-        kubernetes.io/arch: amd64
-        kubernetes.io/os: linux
       {{- end }}
-      {{- if .Values.swoagent.affinity }}
       affinity:
+      {{- if .Values.swoagent.affinity }}
         {{- toYaml .Values.swoagent.affinity | nindent 8 }}
+      {{- else }}
```
pstranak-sw: This is technically a breaking change as setting custom affinity will now effectively remove the platform requirement. But that's the patter we use for other deployed workloads so I kept it here.

PR #1175, opened 2025-10-29
Added state events test (using Clickhouse as storage of timeseries-mock)

gantrior:
* introduced `sw.ebpf.instrumentation/exclude` which will exclude some workloads from beyla autoinstrumentation (e.g. clickhouse cause beyla to crash)
* set filestorage as opt-in (by default disabled). Clickhouse storage is default now.
* Configured Clickhouse to have 5min retention, store all logs, metrics, traces. Use 3GB memory limit.
* Created `generate_relationships.py` which checks current state of the storage and produce asserts based on that - should simplify adding new tests.


deploy/helm/templates/beyla/configmap.yaml
```diff
@@ -39,6 +39,8 @@ data:
         - k8s_namespace: "*"
       exclude_instrument:
         - exe_path: "{*otelcol*|*beyla*}"
+        - k8s_pod_annotations:
+            sw.ebpf.instrumentation/exclude: "true"
```
pstranak-sw: Do we need this annotation? Should it be annotation or label? Is exclude_instrument working as expected here?

doc/development.md
```diff
@@ -221,10 +222,14 @@ helm unittest -u deploy/helm
 To enable code completion when writing new tests, install a VS Code extension providing a YAML Language server, like `redhat.vscode-yaml`.
 
 ## Integration tests
-Integration tests are located in `tests/integration` and are supposed to verify if metric processing is delivering expected outcome.
+Integration tests are located in `tests/integration` and are supposed to verify if metric processing is delivering expected outcome. 
+
+### Timeseries Mock Storage
+Important part of integration test is storage of telemetry data in ClickHouse. It is deployed as part of `skaffold dev`. In order to manually look at the data stored in the Clickhouse use Copilot Prompt `/query-clickhouse` (e.g. `/query-clickhouse show me what relationship state event types were ingested`).
```
pstranak-sw: I'd also mention that for running queries manually, one can use the ClickHouse web UI available at `http://localhost:8123/play`.

tests/integration/generate_relationships.py
```diff
@@ -0,0 +1,496 @@
...
+
+def get_workload_name(entity_id: Dict[str, str]) -> str:
+    for key in K8S_WORKLOAD_KEYS:
+        if key in entity_id:
+            return entity_id[key]
+    return ""
+
+
+def _check_pod_hash_pattern(pod_name: str) -> bool:
+    """Returns True if pod name matches: <base>-<8-10chars>-<5chars> pattern.
```
pstranak-sw: The actual logic can be more complicated. E.g. if the `<base>-<8-10chars>-` is too long, the resulting Pod name will end with `<5chars>` preceded by `<base>-<8-10chars>-` trimmed from the right to fit.

skaffold.yaml
```diff
@@ -33,6 +36,8 @@ deploy:
             uid: '{{ env "CLUSTER_NAME" | default "cluster-uid-123456789" }}'
           otel:
             endpoint: timeseries-mock-service:9082
+            swi_endpoint_check: 
+              enabled: false
```
pstranak-sw: Why?

skaffold.yaml
```diff
@@ -255,7 +266,7 @@ profiles:
   - name: no-tests
     patches:
       - op: remove
-        path: /deploy/helm/releases/3
+        path: /deploy/helm/releases/4
```
pstranak-sw: What is this removing?

tests/integration/run_generate_relationships.sh
pstranak-sw: This script is not mentioned anywhere in the documentation. Should it be?

doc/development.md
gantrior: Hm I actually had no idea that this web UI Clickhouse has. I mentioned it there

doc/development.md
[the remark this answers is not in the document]
gantrior: Agree, Updated doc

skaffold.yaml
pstranak-sw: In other words, the startup of `timeseries-mock-service` now takes much longer and the check times out because of it. OK.

tests/integration/run_generate_relationships.sh
pstranak-sw: I don't see it.

gantrior: https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/1175/commits/e6f8501524b1e0e24b63794b3aeb4e6b7f1a5e38

pstranak-sw (approved):

PR #1189, opened 2025-11-07
Update dependencies

pstranak-sw:


deploy/helm/Chart.yaml
```diff
@@ -20,20 +20,20 @@ maintainers:
 dependencies:
   - name: kube-state-metrics
     repository: https://prometheus-community.github.io/helm-charts
-    version: 6.3.0
+    version: 6.4.1
```
pstranak-sw: No change in behavior.

PR #1192, opened 2025-11-10
Update ArtifactHub info and build scripts for releasing Helm chart

pstranak-sw:
* Remove whitelisting of OTEL eBPF images
* Update chart-releaser
* Handle "beta" and "rc" as pre-release Helm charts
* Add support for "rc" builds to the automated Docker image updater


deploy/helm/Chart.yaml
```diff
@@ -39,16 +39,3 @@ dependencies:
 annotations:
   artifacthub.io/prerelease: "false"
   artifacthub.io/license: Apache-2.0
-  artifacthub.io/images: |
```
pstranak-sw: This is not needed anymore and it disables automatic scan of images in ArtifactHub.

PR #1194, opened 2025-11-12
Switch from the temporary usage of Beyla to OBI, now that it's out

pstranak-sw:


tests/deploy/tests/templates/testResources.yaml
```diff
@@ -1172,10 +1148,9 @@ metadata:
   labels:
     app: external-caller-solarwinds
     component: external-caller
-    test-type: beyla-external-connections
+    test-type: obi-external-connections
   annotations:
     test-annotation: "test-value"
-    beyla.grafana.com/instrument: "true"
```
pstranak-sw: I could not find any documentation about this, nor its counterpart for OBI.

deploy/helm/templates/obi/configmap.yaml
pstranak-sw: While `application_process` should work according to the documentation, the metrics are not produced. Maybe we should just disable it for now?

gantrior: it is probably not yet implemented, because https://opentelemetry.io/docs/zero-code/obi/metrics/ does not include process related metrics. Anyway we don't need them so I would remove this option.

deploy/helm/templates/beyla/configmap.yaml
```diff
@@ -5,30 +5,26 @@
...
-      features: ["application", "application_process", "application_service_graph", "network"]
+      features: ["application", "application_service_graph", "network"]
     {{- end }}
     {{- if not .Values.network_topology.config.ebpf }}
     ebpf:
       context_propagation: all
     {{- end }}
-    {{- if not .Values.network_topology.config.process }}
-    process:
-      enabled: true
```
pstranak-sw: I could not find any information about this setting (including searching in the source code).

PR #1314, opened 2026-04-02
updating istio metrics filters

mpecenka:
updating istio metrics filters to be mutually exclusive ie. istio datapoints should be exported just once


etichy (approved):
We don't cover scenario where workload and service are from different namespace. But this change improve situation. Also would be nice to switch to router to reduce overhead with cloning datapoints.

PR #1316, opened 2026-04-08
istio metrics fix

mpecenka:
delete duplicated istio metrics before those are exported (events are still generated)


pstranak-sw (approved):
This will work, but it will be hard to understand what is happening here and why.

PR #1335, opened 2026-05-18
Add EntityRefs to Container logs

pstranak-sw:
It will make entity inference more reliable than when just using Resource attributes.

Depends on https://github.com/solarwinds/solarwinds-otel-collector-contrib/pull/318

The `EntityRefs` section of the OTEL message is not exposed through the integration test database, so there is no integration test for this.

### DIFF
--- a/deploy/helm/tests/__snapshot__/node-collector-config-map-windows_test.yaml.snap
+++ b/deploy/helm/tests/__snapshot__/node-collector-config-map-windows_test.yaml.snap
@@ -1355,6 +1355,11 @@ Node collector config for windows nodes should match snapshot when using default
           - id_keys:
             - sw.k8s.cluster.uid
             - k8s.namespace.name
+            - k8s.replicaset.name
+            type: KubernetesReplicaSet
+          - id_keys:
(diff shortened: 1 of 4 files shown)


deploy/helm/tests/__snapshot__/node-collector-config-map-windows_test.yaml.snap
```diff
@@ -1329,6 +1329,49 @@ Node collector config for windows nodes should match snapshot when using default
...
+            - sw.k8s.cluster.uid
+            - k8s.namespace.name
+            - k8s.job.name
+            type: KubernetesJob
+          - id_keys:
+            - sw.k8s.cluster.uid
+            - k8s.namespace.name
+            - k8s.cronjob.name
+            type: KubernetesCronJob
         transform/istio-metric-datapoints:
```
mpecenka, line 1380: no logs for KubernetesReplicaSet?

pstranak-sw: added

mpecenka (approved):

PR #1340, opened 2026-05-26
Add entity discovery based on running processes

pstranak-sw:
Add discovery of processes that could be instrumented by APM.

Depends on new `swotelcol` release.


deploy/helm/node-collector-config-entity-discovery.yaml
```diff
@@ -0,0 +1,180 @@
+{{- if .Values.entityDiscovery.enabled }}
+receivers:
+  hostmetrics/entity-discovery:
+    collection_interval: 30s
```
mpecenka, line 4: i would increase the interval to match other polling

PR #1349, opened 2026-06-10
Show error when deploying trivy-operator to OpenShift cluster

pstranak-sw:
It's very likely that the user needs to configure image pull secrets to be able to scan most of the deployed images. And it also causes issues with upgrades.

So we are disabling this option.

### DIFF
--- a/deploy/helm/templates/NOTES.txt
+++ b/deploy/helm/templates/NOTES.txt
@@ -163,7 +163,7 @@ WARNING: Deploying Aqua Security Trivy-Operator. SolarWinds does not provide any
 {{- end -}}
 
 
-{{- if index .Values.ebpfNetworkMonitoring.enabled -}}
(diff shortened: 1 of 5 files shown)


deploy/helm/templates/NOTES.txt
```diff
@@ -163,7 +163,12 @@ WARNING: Deploying Aqua Security Trivy-Operator. SolarWinds does not provide any
 {{- end -}}
 
 
-{{- if index .Values.ebpfNetworkMonitoring.enabled -}}
+{{- if .Values.ebpfNetworkMonitoring.enabled -}}
```
pstranak-sw, line 166: This was probably a typo, working correctly by accident.

juraj-paska (dismissed):

juraj-paska (approved):

PR #1350, opened 2026-06-11
fix: rewrite automated image updater and unblock its PR pipeline

murad-sw:
## Summary

The automated Docker image updater workflow has been failing for months because `updateImages.yml` referenced `.github/scripts/update_docker_images.py` and `.github/scripts/requirements.txt`, neither of which exists in the repo (the script lives at `utils/update_images.py` with no requirements file).

- **Rewritten updater** at `.github/update-images/`:
  - Correct registry routing for Docker Hub, GHCR, Quay, gcr.io, `*.pkg.dev`, and arbitrary v2 registries (the previous code mis-parsed `solarwinds/swo-agent` as host=`solarwinds` and queried `library/swo-agent` on Docker Hub).
  - Prerelease policy: skip prerelease tags unless the current pinned tag is itself a prerelease (so stable images don't get bumped to RCs).
- **41 unit tests** under `.github/update-images/test_update_images.py`, gated as a prerequisite job in `updateImages.yml`.
- **Decoupled helm unit tests from values.yaml-pinned image tags.


.github/update-images/config.yaml
```diff
@@ -0,0 +1,24 @@
...
+#                           pinned tag is itself a prerelease
+#       pin_major:          int; only consider versions with this major
+#       pin_minor:          int; only consider versions with this major.minor
+#       tag_prefix:         str; only consider tags starting with this prefix
+#                           (e.g. "v"). Empty string means strip-`v` allowed.
+
+ignore:
+  # Beyla pulls in eBPF kernel collectors that need version-locked support.
+  # Bumps must be coordinated across multiple images, so we don't auto-update.
+  - grafana/beyla
```
pstranak-sw, line 17: I'm not sure what this means. There is only one image. Also, we use OBI (`opentelemetry-ebpf-instrumentation/ebpf-instrument`), bot Beyla.

.github/update-images/config.yaml
```diff
@@ -0,0 +1,24 @@
...
+  - solarwinds/opentelemetry-ebpf-kernel-collector
+  - solarwinds/opentelemetry-ebpf-k8s-watcher
+  - solarwinds/opentelemetry-ebpf-k8s-relay
+  - solarwinds/opentelemetry-ebpf-reducer
```
pstranak-sw, line 21: BTW, we will be removing these from the chart in the near future.

deploy/helm/tests/swo-agent-statefulset_test.yaml
```diff
@@ -7,6 +7,7 @@ tests:
     template: swo-agent-statefulset.yaml
     set:
       swoagent.enabled: true
+      swoagent.image.tag: "test-tag"
```
pstranak-sw, line 10: We already do this for tests that reference the main image. So I think we can do it for other images, like these, too. In case we this PR does not get merged, these changes could still be merged in a separate PR.

PR #1358, opened 2026-06-19
disabled Logs collection bydefault

etichy:
Logs collection will be disabled by default. Existing deployments will stay without change


deploy/helm/README.md
```diff
@@ -314,7 +314,7 @@ Configuration:
 
 ### Logs Collection
 
-Enabled by default. To disable, set `otel.logs.enabled: false`.
+Disabled by default. To enabled, set `otel.logs.enabled: true`.
```
pstranak-sw: I'd include the info that is a change starting with some version (5.3.0 ?)

deploy/helm/templates/_helpers.tpl
```diff
@@ -424,6 +424,32 @@ Output will be the same format but tripled
...
+resolved effective value, so the correct state is carried forward across all upgrades.
+
+Usage:
+{{- if include "isLogsEnabled" . }}
+*/}}
+{{- define "isLogsEnabled" -}}
+{{- if kindIs "bool" .Values.otel.logs.enabled -}}
+  {{- if .Values.otel.logs.enabled -}}true{{- end -}}
+{{- else -}}
+  {{- $existingCM := lookup "v1" "ConfigMap" .Release.Namespace (include "common.fullname" (tuple . "-node-collector-config")) -}}
```
pstranak-sw, line 445: `lookup` is not allowed for AWS addons.

deploy/helm/tests/__snapshot__/node-collector-config-map-windows_test.yaml.snap
```diff
@@ -1699,27 +1698,6 @@ Node collector config for windows nodes should match snapshot when using default
         - health_check
         - k8s_observer
         pipelines:
-          logs:
```
pstranak-sw, line 1702: Add a test for explicitly enabled/disabled/undefined logs. Since the default is changing, we are loosing test coverage if we don't add new tests.

deploy/helm/templates/_helpers.tpl
etichy: this behaves same as for helm template where API is not available. It returns empty result set =  we will always disable logs on upgrade for these customers.

etichy: It's not ideal, but I don't see other way and we have to break users of EKS addon here.

pstranak-sw (approved):

PR #1375, opened 2026-07-09
Fix memory and cpu metrics on windows nodes

etichy:
cAdvisor is not avilable on windows nodes. We switched to kubeletstats on windows nodes for now.

### DIFF
--- a/deploy/helm/node-collector-config.yaml
+++ b/deploy/helm/node-collector-config.yaml
@@ -122,6 +122,44 @@ processors:
       - k8s.node.cpu.usage.seconds.rate
 {{- end }}
 
+{{- if and (not .isWindows) .Values.otel.metrics.kubeletstats.enabled .Values.otel.metrics.enabled (not .Values.aws_fargate.enabled) }}
+  # Drop CPU metrics from cAdvisor pipeline to avoid duplicates when kubeletstats is collecting them.
+  filter/cadvisor-cpu:
+    error_mode: ignore
+    metrics:
+      metric:
+        - name == "k8s.node.cpu.usage.seconds.rate"
+        - name == "k8s.pod.cpu.usage.seconds.rate"
+        - name == "k8s.container.cpu.usage.seconds.rate"
+
+  metrics_transform/linux-kubeletstats:
+    transforms:
+      - include: k8s.container.cpu.time
+        action: update
+        new_name: k8s.container.cpu.usage.seconds.rate
+      - include: k8s.pod.cpu.time
+        action: update
+        new_name: k8s.pod.cpu.usage.seconds.rate
+      - include: k8s.node.cpu.time
+        action: update
+        new_name: k8s.node.cpu.usage.seconds.rate
+
+  cumulativetodelta/linux-kubeletstats:
+    max_staleness: {{ include "common.maxStaleness" . }}
+    include:
+      metrics:
+        - k8s.container.cpu.usage.seconds.rate
+        - k8s.pod.cpu.usage.seconds.rate
+        - k8s.node.cpu.usage.seconds.rate
+      match_type: strict
+
+  deltatorate/linux-kubeletstats:
+    metrics:
+      - k8s.container.cpu.usage.seconds.rate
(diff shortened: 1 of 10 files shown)


deploy/helm/node-collector-config.yaml
```diff
@@ -750,6 +849,32 @@ service:
       exporters:
         - forward/metric-exporter
 {{- end }}
+{{- if and .isWindows .Values.otel.metrics.enabled (not .Values.aws_fargate.enabled) }}
+    metrics/node-windows:
```
gantrior, line 853: Why this pipeline skips `resource/metrics`?

deploy/helm/templates/cluster-role.yaml
```diff
@@ -18,6 +18,7 @@ rules:
       - nodes/spec
       - nodes/proxy
       - nodes/metrics
+      - nodes/stats
```
gantrior, line 21: if there is Helm condition for kubeletstats, this role here could be conditioned as well

gantrior (changes requested):
The PR claims to fix the memory as well, but does touch only CPU related metrics.

The telemetry is never tested in integration tests so it will be very fragile. I do not like having kubeletstats just for windows, because I assume the telemetry of cAdvisor (on Windows) and kubeletstats (on Linux) will not be equivalent. 

I think this issue should be broader and should replace cAdvisor with kubeletstats generally (which will have UI and migration implications).

deploy/helm/node-collector-config.yaml
pstranak-sw: Many of the attributes are on the metrics already. Maybe all?

deploy/helm/node-collector-config.yaml
```diff
@@ -750,6 +849,32 @@ service:
...
+        - deltatorate/windows-kubeletstats
+        - resource/all
+      exporters:
+        - forward/metric-exporter
+{{- end }}
+{{- if and (not .isWindows) .Values.otel.metrics.kubeletstats.enabled .Values.otel.metrics.enabled (not .Values.aws_fargate.enabled) }}
+    metrics/node-kubeletstats:
+      receivers:
+        - kubelet_stats/linux
+      processors:
```
pstranak-sw, line 869: If this is about the CPU metrics right now, shouldn't we drop the other we receive from this receiver? Or even better, disable their collection in the receiver itself?

deploy/helm/node-collector-config.yaml
```diff
@@ -658,6 +728,32 @@ receivers:
...
+    endpoint: "https://${NODE_IP}:10250"
+    insecure_skip_verify: true
+    metric_groups:
+      - container
+      - pod
+      - node
+{{- end }}
+
+{{- if and (not .isWindows) .Values.otel.metrics.kubeletstats.enabled .Values.otel.metrics.enabled (not .Values.aws_fargate.enabled) }}
+  kubelet_stats/linux:
```
pstranak-sw, line 746: Do we need also this?
```yaml
extra_metadata_labels:
  - container.id
 ```
Though that would need also `nodes/pods` permission.

File: deploy/helm/templates/node-collector-daemon-set.yaml [hunk 5]
```diff
@@ -138,6 +138,10 @@ spec:
               valueFrom:
                 fieldRef:
                   fieldPath: spec.nodeName
+            - name: NODE_IP
```
pstranak-sw, line 141: The receiver's documentation mentions `NODE_NAME`, not `NODE_IP`. Why this one?

deploy/helm/node-collector-config.yaml
```diff
@@ -92,6 +92,74 @@ processors:
     metrics:
       {{- include "common-config.cumulativetorate-cadvisor" . | nindent 6}}
 
+{{- if and .isWindows .Values.otel.metrics.enabled (not .Values.aws_fargate.enabled) }}
```
pstranak-sw, line 95: Why have these processors separate between Windows and Linux, when they are the same?

PR #1376, opened 2026-07-13
Add tailsampling of traces to gateway

pstranak-sw:
Introduce tail sampling configuration for traces in the gateway collector. This changes the gateway collector from Deployment to StatefulSet.

The default tailsampling configuration is based on https://github.com/solarwinds/solarwinds-otel-collector-releases/blob/d565145d2a494c91f1a93a3033daee9c40e4aa5f/examples/tail-sampling/collector.tail-sampling.yaml. Except it does not contain the `critical-transactions-policy` (which needs to be set specifically for customer's environment) and it uses cache sizes that are closer to the default values.

Additionally, explicitly add Kubernetes cluster attributes to internal telemetry.

This is a rework of https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/1293.


deploy/helm/values.yaml
```diff
@@ -854,10 +854,101 @@ otel:
...
+
+      # Number of traces to buffer for sampling decisions.
+      # Should be at least: expected_new_traces_per_sec × decision_wait (in seconds)
+      # Example: 500 traces/sec × 10s = 5000 (minimum buffer size)
+      # Increase this value if your trace throughput exceeds expected_new_traces_per_sec.
+      num_traces: 50000
+
+      # Expected rate of new traces per second.
+      # Used together with num_traces to size the in-memory buffer.
+      expected_new_traces_per_sec: 500
```
pstranak-sw, line 879: If these numbers are too low, all traces may get dropped.

deploy/helm/templates/gateway/gateway-statefulset.yaml
```diff
@@ -1,6 +1,6 @@
 {{- if .Values.otel.gateway.enabled }}
 apiVersion: apps/v1
-kind: Deployment
+kind: StatefulSet
```
gantrior, line 3: `kind` is immutable in kubernetes, shouldn't we change also the name to not leave orphaned Deployment alongside a StatefulSet during upgrade? Did you test the upgrade?

deploy/helm/templates/gateway/gateway-statefulset.yaml
gantrior: This only hashes config, not the config-traces.yaml

deploy/helm/templates/gateway/gateway-config-map.yaml
```diff
@@ -11,5 +11,7 @@ metadata:
 {{ include "common.annotations" . | indent 4 }}
 data:
   gateway.config: |
-{{ tpl (.Files.Get "gateway-collector-config.yaml") . | fromYaml | toYaml | indent 8 }}
+{{- $main := tpl (.Files.Get "gateway-collector-config.yaml") . | fromYaml }}
+{{- $traces := tpl (.Files.Get "gateway-collector-config-traces.yaml") . | fromYaml }}
+{{ merge $main $traces | toYaml | indent 8 }}
```
gantrior: I am afraid of using `merge` will be fragile, as it favors $main on any future key collision with no error. What was the reason to choose merge instead of keeping whole config in single file?

deploy/helm/gateway-collector-config-traces.yaml
```diff
@@ -0,0 +1,125 @@
...
+# gateway-collector-config.yaml (currently: all traces/* pipelines are owned here).
+
+exporters:
+{{- if .Values.otel.gateway.tailSampling.enabled }}
+  # Load balancing exporter for tail-sampling
+  # Routes traces by traceID to specific gateway replicas via headless service
+  loadbalancing:
+    routing_key: "traceID"
+    protocol:
+      otlp:
```
gantrior, line 12: Why there is no `retry_on_failure` like in other `otlp` related exporters?

deploy/helm/values.yaml
```diff
@@ -854,10 +854,101 @@ otel:
...
+
+      # Decision wait time - how long to wait before making a sampling decision
+      # This allows time for all spans of a trace to arrive
+      decision_wait: 10s
+
+      # Number of traces to buffer for sampling decisions.
+      # Should be at least: expected_new_traces_per_sec × decision_wait (in seconds)
+      # Example: 500 traces/sec × 10s = 5000 (minimum buffer size)
+      # Increase this value if your trace throughput exceeds expected_new_traces_per_sec.
+      num_traces: 50000
```
gantrior: Shouldn't we increase gateway memory requests/limits (1Gi/400Mi) due to this increase trace buffer?

gantrior (changes requested):

deploy/helm/templates/gateway/gateway-statefulset.yaml
pstranak-sw: Helm doesn't change the kind. It destroys the old object and creates the new one.

deploy/helm/templates/gateway/gateway-statefulset.yaml
pstranak-sw: Good catch. Fixed.

deploy/helm/templates/gateway/gateway-config-map.yaml
pstranak-sw: Readability reasons. The config files are big and contain unrelated pipelines. Splitting them helps me with filtering out unrelated stuff when looking into them.
Also, this was proposed in my older PR https://github.com/solarwinds/swi-k8s-opentelemetry-collector/pull/1340, which is postponed for now for unrelated reasons.

pstranak-sw: The merging concern will be an issue only if we use the same names in both files for the same kinds of objects (processor/connector/...). And in that rare case, the developer should notice that the new code is not working as expected.

deploy/helm/gateway-collector-config-traces.yaml
pstranak-sw: I added one now, following the pattern of the queue options. Though, I'm not sure we need it. And reusing the main gateway's settings is not ideal, as intra-cluster communication between the collector's parts should be much quicker than connections to the Internet. We'll see how it goes.

deploy/helm/values.yaml
pstranak-sw: I decreased the buffer a bit, as I don't expect so many traces will be needed to be kept in memory, since we have HPA for the gateway collector. And that should take care of the memory limits, too.

deploy/helm/templates/gateway/gateway-config-map.yaml
gantrior: OTEL collector itself also has merge capabilities as it can basically accept multiple configs, but I am not sure if it is any better. Lets rely on snapshot test/integration test to catch the future collision. Accepting...

gantrior (approved):

PR #1387, opened 2026-08-03
Swotelcol entities per collector instance

etichy:
Publish swotelcol entities per deployment instance.

### DIFF
--- a/deploy/helm/templates/gateway/gateway-statefulset.yaml
+++ b/deploy/helm/templates/gateway/gateway-statefulset.yaml
@@ -118,9 +118,13 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: status.podIP
+            - name: POD_NAME
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
 {{- if .Values.otel.swotelcol_entities.enabled }}
             - name: SWOTELCOL_COLLECTOR_NAME
-              value: "$(CLUSTER_NAME)-gateway"
+              value: "$(CLUSTER_NAME)-$(POD_NAME)"
 {{- end }}
             - name: SOLARWINDS_API_TOKEN
               valueFrom:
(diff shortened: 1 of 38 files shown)


deploy/helm/templates/gateway/gateway-statefulset.yaml
```diff
@@ -118,6 +118,18 @@ spec:
                 fieldRef:
                   apiVersion: v1
                   fieldPath: status.podIP
+            - name: POD_NAME
```
pstranak-sw: Why do we need the `PODNAME`?

deploy/helm/templates/gateway/gateway-statefulset.yaml
```diff
@@ -118,6 +118,18 @@ spec:
...
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.name
+{{- if .Values.otel.swotelcol_entities.enabled }}
+            - name: POD_ORDINAL
+              valueFrom:
+                fieldRef:
+                  fieldPath: metadata.labels['apps.kubernetes.io/pod-index']
+            - name: SWOTELCOL_COLLECTOR_NAME
+              value: "$(CLUSTER_NAME)-gateway-$(POD_ORDINAL)"
```
pstranak-sw, line 127: So we want to have a bit more user-friendly names than simple Pod names, right? I'm fine with what, but we will have to be careful to not make customers believe that they can rely on stability of the name.

deploy/helm/values.yaml
```diff
@@ -772,6 +772,11 @@ otel:
   swi_endpoint_check:
     enabled: true
 
+  # Configuration for SwoTelcol entity creation via solarwindsextension.
+  # Each collector instance registers itself as a SwoTelcol entity in SolarWinds Observability.
+  swotelcol_entities:
```
pstranak-sw, line 777: I'm not sure if it's better here in `otel` or in the root. I guess we can keep this as-is.

deploy/helm/templates/gateway/gateway-statefulset.yaml
etichy: leftover good catch

pstranak-sw (approved):

